PR #7, opened 2022-10-03
PACT-381 - EPR - Rendering Edit Set from model

enriquedelpino:
The Edit Set page now renders a table according to GDS components that takes a EntrySet model object and creates a row in the table iteratively from a sequence of EntrySetEntry objects.

Also the heading of the page is now created from a parametrized message along with the Entry Set name.

### DIFF
--- a/app/uk/gov/nationalarchives/omega/editorial/views/editSet.scala.html
+++ b/app/uk/gov/nationalarchives/omega/editorial/views/editSet.scala.html
@@ -1,5 +1,26 @@
-@(title: String, heading: String)
+@import uk.gov.nationalarchives.omega.editorial.models.{ EditSet, EditSetEntry }
+
+@(title: String, heading: String, editSet: EditSet)
 
 @template(title) {
   <h1>@heading</h1>
+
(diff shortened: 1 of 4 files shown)


rwalpole (changes requested):

app/uk/gov/nationalarchives/omega/editorial/views/editSet.scala.html
```diff
@@ -3,9 +3,8 @@
 @(title: String, heading: String, editSet: EditSet)
 
 @template(title) {
-  <h1>@heading</h1>
-
-  <table class="govuk-table">
+ <table class="govuk-table">
+    <caption class="govuk-table__caption govuk-table__caption--l">@heading</caption>
```
rwalpole, line 7: Can we use `govuk-table__caption--m` please as per the prototype: https://ctd-omega-ui-prototype.herokuapp.com/edit-set

rwalpole (changes requested):

rwalpole (approved):

PR #19, opened 2022-10-19
PACT-449: updating to match prototype

rwalpole:


enriquedelpino (approved):
This looks good to me

PR #22, opened 2022-10-28
PACT-445 | EPR | Loading data from case classes

enriquedelpino:
There is now a package object that contains the data for the edit set records that gets loaded into the forms. This data is indexed using the
 OCI code that will be used to map incoming request and populate the
 form accordingly.

Also the data displayed on the discard and save pages is loaded from the data map and links to return to previous page are fixed.


app/uk/gov/nationalarchives/omega/editorial/models/package.scala
```diff
@@ -0,0 +1,65 @@
...
+    "COAL.2022.V4RJW.P" -> new EditSetRecord(
+      "COAL 80/80/2",
+      "COAL.2022.V4RJW.P",
+      "Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths.",
+      "1960",
+      "",
+      "",
+      ""
+    ),
+    "COAL.2022.V2RJW.P" -> new EditSetRecord(
```
rwalpole, line 46: Should this be `COAL.2022.V3RJW.P`?

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -120,10 +118,13 @@ class EditSetController @Inject() (
...
       logger.info(s"The edit set id is $id for record id $recordId")
+      val record = editSetRecords.getEditSetRecordByOCI(recordId).getOrElse(defaultEditSetRecord)
       val messages: Messages = request.messages
       val title: String = messages("edit-set.record.edit.title")
-      val heading: String = messages("edit-set.record.edit.heading")
-      Ok(editSetRecordEdit(title, heading, editSetRecordForm))
+      val heading: String = messages("edit-set.record.edit.heading", record.ccr)
+
+      val recordForm = editSetRecordForm.fill(record)
+      Ok(editSetRecordEdit(title, heading, recordForm))
```
rwalpole, line 127: I think rather than returning the `defaultEditSetRecord` it would be better to return an HTTP 404 Not Found error. You could do this like so:-
```
editSetRecords.getEditSetRecordByOCI(recordId) match {
    case Some(record) =>
        val messages: Messages = request.messages
        val title: String = messages("edit-set.record.edit.title")
        val heading: String = messages("edit-set.record.edit.heading", record.ccr)
        val recordForm = editSetRecordForm.fill(record)
        Ok(editSetRecordEdit(title, heading, recordForm))
    case None => NotFound
}
```

rwalpole (changes requested):

PR #27, opened 2022-11-04
PACT-421 | EPR | Hardcoded models replaced by json

enriquedelpino:
EdiSet and EditSetRecord case clases are now populated and read from json strings.

### DIFF
--- a/app/uk/gov/nationalarchives/omega/editorial/models/EditSet.scala
+++ b/app/uk/gov/nationalarchives/omega/editorial/models/EditSet.scala
@@ -38,34 +38,12 @@ case class EditSetRecord(
   endDate: String
 )
 object EditSetRecord {
-  implicit val editSetRecordReads: Reads[EditSetRecord] =
-    (
-      (JsPath \ "ccr").read[String](minLength[String](5)) and
-        (JsPath \ "oci").read[String](minLength[String](5)) and
-        (JsPath \ "scopeAndContent").read[String](minLength[String](10)) and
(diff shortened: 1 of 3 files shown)


app/uk/gov/nationalarchives/omega/editorial/models/EditSet.scala
```diff
@@ -33,3 +37,35 @@ case class EditSetRecord(
...
+    (
+      (JsPath \ "ccr").read[String](minLength[String](5)) and
+        (JsPath \ "oci").read[String](minLength[String](5)) and
+        (JsPath \ "scopeAndContent").read[String](minLength[String](10)) and
+        (JsPath \ "coveringDates").read[String](minLength[String](0)) and
+        (JsPath \ "formerReferenceDepartment").read[String](minLength[String](0)) and
+        (JsPath \ "startDate").read[String](minLength[String](0)) and
+        (JsPath \ "endDate").read[String](minLength[String](0))
+    )(EditSetRecord.apply _)
+}
```
rwalpole, line 51: You can just do `implicit val editSetRecordReads: Reads[EditSetRecord] = Json.reads[EditSetRecord]` but I see you've added some validation to these fields. We need to be a bit careful about this, for example scope and content can be empty. Also I don't think `minLength[String](0)` would have any effect. We also need to think carefully about where we do validation as if we do it on more than one place we could end up having different values acceptable in different places. Let's discuss..

rwalpole (approved):

PR #35, opened 2022-12-12
PACT-590 Added second set of submit buttons to edit record screen.

tna-erasmos:
I also added a few more custom matchers, which are more aware of the view structure.

### DIFF
--- a/app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
+++ b/app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
@@ -20,110 +20,108 @@
  *@
 
 @import helper._
-@import uk.gov.nationalarchives.omega.editorial.models.{ EditSetRecord, User }
-@import uk.gov.nationalarchives.omega.editorial.forms.GovukHelpers._
 @import uk.gov.hmrc.govukfrontend.views.html.components._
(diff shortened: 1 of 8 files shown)


app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -20,110 +20,108 @@
...
+    </div>
 }
 
 @fieldsetContent = {
-  @CSRF.formField
-  @govukTextarea(Textarea(
+@CSRF.formField
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+        <h3 class="govuk-fieldset__heading">@messages("edit-set.record.edit.legend")</h3>
+    </legend>
```
rwalpole, line 54: I would prefer that we use the view models as far as possible, in this case the [Legend view model](https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/scala/uk/gov/hmrc/govukfrontend/views/viewmodels/fieldset/Legend.scala). This is how it was done previously.

tna-erasmos: OK. I'll try that.

tna-erasmos: Updated.

rwalpole (approved):

PR #37, opened 2022-12-14
PACT 625 Fix bug with form validation occuring on discard in edit record page

felixgb:


test/controllers/EditSetControllerSpec.scala
```diff
@@ -306,6 +306,33 @@ class EditSetControllerSpec extends BaseSpec {
 
       status(editRecordPage) mustBe SEE_OTHER
     }
+
+    "redirect to discard page on form submit even if validation fails" in {
+      val blankScopeAndContentToFailValidation = ""
```
tna-erasmos, line 311: Good. Makes it clear.

test/controllers/EditSetControllerSpec.scala
```diff
@@ -306,6 +306,33 @@ class EditSetControllerSpec extends BaseSpec {
...
+          .withSession(SessionKeys.token -> validSessionToken)
+      )
+      val editRecordPage = route(app, request).get
+
+      status(editRecordPage) mustBe SEE_OTHER
+
+      val expectedRedirectLocation = "/edit-set/1/record/COAL.2022.V5RJW.P/edit/discard"
+      val Some(discardPage) = redirectLocation(editRecordPage)
+
+      expectedRedirectLocation mustBe discardPage
```
tna-erasmos, line 333: Or perhaps:

```
redirectLocation(editRecordPage) mustBe Some("/edit-set/1/record/COAL.2022.V5RJW.P/edit/discard")
```

test/controllers/EditSetControllerSpec.scala
```diff
@@ -306,6 +306,33 @@ class EditSetControllerSpec extends BaseSpec {
 
       status(editRecordPage) mustBe SEE_OTHER
     }
+
+    "redirect to discard page on form submit even if validation fails" in {
+      val blankScopeAndContentToFailValidation = ""
+      val request = CSRFTokenHelper.addCSRFToken(
+        FakeRequest(POST, "/edit-set/1/record/COAL.2022.V5RJW.P/edit")
```
tna-erasmos, line 313: This is out-of-scope, but just an observation: we probably should assert that the record hasn't been changed.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -153,4 +156,7 @@ class EditSetController @Inject() (
     }
   }
 
+  private def formToEither[A](form: Form[A]): Either[Form[A], A] =
```
tna-erasmos, line 159: This would be useful in other places. Should we put it somewhere common now, or wait until another needs it? The other issue with the latter, is that they may not be aware of it.

tna-erasmos (commented):
Merely suggestions.

test/controllers/EditSetControllerSpec.scala
felixgb: I agree; made the chage. Still under 120 characters.

test/controllers/EditSetControllerSpec.scala
felixgb: Hm, I agree that this kind of test would be useful, however I don't see any spec that tries to model that behavior (or even a spec that asserts behavior across pages).

Maybe we could bring up creating some more specs at the very highest level, e.g.
log in -> edit a set A -> discard -> assert there are no chages on A

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
felixgb: I'm wary of creating "utility" files full of helper methods at this point where they are only used in one place, however as soon as it's used in another place we should put it in a controllers package object, I think.

I think we can just flag it in code review if it come up?

test/controllers/EditSetControllerSpec.scala

test/controllers/EditSetControllerSpec.scala
tna-erasmos: Yes - as I wrote, it's out-of-scope. Just something to keep in mind going forward.

It can wait for awhile.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: OK.

tna-erasmos (approved):
LGTM

PR #38, opened 2022-12-15
PACT-619 Internationalize all strings

felixgb:


conf/messages
```diff
@@ -1,35 +1,44 @@
...
+login.authentication.problem=There is a problem
 login.button.text=Sign in
 login.heading=Sign in: Pan-Archival Catalogue
 login.missing.password=Enter a password
 login.missing.username=Enter a username
 login.password.label=Password
 login.title=Sign in
 login.username.label=Username
+
+placeholder.back-to-coal-80=Back to edit set (COAL 80 Sample)
```
felixgb, line 42: Added these placeholders into messages as well, I think it will make them easier to delete

app/uk/gov/nationalarchives/omega/editorial/views/editSet.scala.html
```diff
@@ -53,13 +53,13 @@
       head = Some(
           Seq(
               HeadCell(
-                  content = Text("CCR")
+                  content = Text(messages("edit-set.record.edit.ccr"))
```
tna-erasmos, line 56: (Refers to line 28)

This might be out of scope, but: Why are we sometimes passing in messages that we could resolve within the view?

tna-erasmos:
You missed "Error:" in [govukLoginInput.scala.html.](https://github.com/nationalarchives/ctd-omega-editorial-frontend/blob/PACT-619/app/uk/gov/nationalarchives/omega/editorial/views/govukLoginInput.scala.html#L27)

conf/messages
tna-erasmos: I think the edit-set-specific bit (e.g. "COAL 80 Sample") could be passed into the view. It's just the edit set name, after all.

tna-erasmos:
Also, one missed in [editSetRecordEdit.scala.html](https://github.com/nationalarchives/ctd-omega-editorial-frontend/blob/PACT-619/app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html#L120).

> There is a problem.

The key is  `error.summary.title`.

app/uk/gov/nationalarchives/omega/editorial/views/editSet.scala.html
felixgb: I agree that we should be resolving messages within the views, however I think this would require a lot of refactoring to do consistently and would be out of scope.

tna-erasmos: Agreed - one day.

conf/messages
tna-erasmos: OK.

tna-erasmos (approved):
LGTM

PR #42, opened 2022-12-20
PACT-640: Added Legal Status to the edit record page

userjd:


test/views/EditRecordViewSpec.scala
```diff
@@ -179,15 +202,16 @@ class EditRecordViewSpec extends BaseSpec {
         startDateYear = "1960",
         endDateDay = "31",
         endDateMonth = "12",
-        endDateYear = "1960"
+        endDateYear = "1960",
+        legalStatus = "1234"
```
tna-erasmos, line 206: Could you make this realistic and valid?

test/support/CustomMatchers.scala
```diff
@@ -204,6 +204,26 @@ object CustomMatchers {
...
   def haveNoErrorMessageForCoveringDate: Matcher[Document] = haveErrorMessageForCoveringDate("")
 
+  def haveErrorMessageForLegalStatus(expectedValue: String): Matcher[Document] = (document: Document) =>
+    singleValueMatcher(
+      "an error message for legal status",
+      expectedValue,
+      document.select("#legalStatus-error").text()
+    )
+
+  def haveLegalStatus(expectedValue: String): Matcher[Document] = (document: Document) => {
```
tna-erasmos, line 214: This could also use 'singleValueMatcher'.

test/support/BaseSpec.scala
```diff
@@ -38,4 +38,13 @@ class BaseSpec extends PlaySpec with GuiceOneAppPerSuite with Injecting {
 
   def asDocument(rawContent: String): Document = Jsoup.parse(rawContent)
 
+  val legalStatusData =
```
tna-erasmos, line 41: Given that this is only used in the one spec `EditRecordViewSpec` I would suggest that it goes there, instead.

test/controllers/EditSetControllerSpec.scala
```diff
@@ -660,6 +680,7 @@ class EditSetControllerSpec extends BaseSpec {
                 "endDateDay"                -> "31",
                 "endDateMonth"              -> "12",
                 "endDateYear"               -> "2020",
+                "legalStatus"               -> "1234",
```
tna-erasmos, line 683: Could you make this realistic and valid?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -618,6 +637,7 @@ class EditSetControllerSpec extends BaseSpec {
...
```
tna-erasmos, line 640: Could you make this realistic and valid?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -563,6 +580,7 @@ class EditSetControllerSpec extends BaseSpec {
                       "endDateDay"                -> "31",
                       "endDateMonth"              -> "12",
                       "endDateYear"               -> "2020",
+                      "legalStatus"               -> "1234",
```
tna-erasmos, line 583: Could you make this realistic and valid?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -516,6 +519,20 @@ class EditSetControllerSpec extends BaseSpec {
           }
         }
 
+        "legal status" when {
+          "is not selected" in {
+
+            val values = validValues ++ Map("legalStatus" -> "ref.0")
```
tna-erasmos, line 525: I would make this blank.

conf/messages
```diff
@@ -12,13 +12,15 @@ edit-set.record.edit.former-reference=Former reference - Department (optional)
...
 edit-set.record.edit.legend=Intellectual properties
+edit-set.record.edit.legal-status=Legal Status
 edit-set.record.edit.scope-and-content=Scope and content
 edit-set.record.edit.start-date=Start date
 edit-set.record.edit.title=Edit record
 edit-set.record.error.end-date-before-start-date=End date cannot precede start date
 edit-set.record.error.end-date=End date is not a valid date
 edit-set.record.error.former-reference-department=Former reference - Department too long, maximum length 255 characters
 edit-set.record.error.scope-and-content=Scope and content too long, maximum length 8000 characters
+edit-set.record.error.legal-status=Select a valid legal status
```
tna-erasmos, line 23: Could you keep this list sorted?

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -147,6 +151,26 @@
         html = dateSelector("end", "edit-set.record.edit.end-date"),
         role = Some("group")
     ))
+@govukSelect(Select(
+    id = "legalStatus",
+    name = "legalStatus",
+    items = legalStatusData.map {status =>
```
tna-erasmos, line 157: This is where I'd put the blank option, with a message lookup.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -22,11 +22,15 @@
...
+@import uk.gov.nationalarchives.omega.editorial.models.{EditSetRecord, User, LegalStatus}
+@import uk.gov.hmrc.govukfrontend.views.html.components.implicits._
 
-@this(govukButton: GovukButton, govukInput: GovukInput, govukTextarea: GovukTextarea, govukFieldset: GovukFieldset, govukErrorSummary: GovukErrorSummary)
 
-@(user: User, title: String, editSetRecordForm: Form[EditSetRecord])(implicit messages: Messages, request: Request[AnyContent])
+
+@this(govukButton: GovukButton, govukInput: GovukInput, govukTextarea: GovukTextarea, govukFieldset: GovukFieldset, govukErrorSummary: GovukErrorSummary, govukSelect: GovukSelect)
+
+@(user: User, title: String, legalStatusData: Seq[LegalStatus], editSetRecordForm: Form[EditSetRecord])(implicit messages: Messages, request: Request[AnyContent])
```
tna-erasmos, line 32: Perhaps rename to `legalStatuses`, as `data` is a bit generic.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
```diff
@@ -133,3 +136,15 @@ package object editSets {
   def getEditSet(): EditSet =
     editSet1
 }
+
+package object legalStatus {
+  val legalStatusData =
+    Seq(
+      LegalStatus("ref.0", "Select a Legal Status"),
```
tna-erasmos, line 143: This makes it look like `ref.0` is a legitimate legal service; instead I would remove this element, and add the initial element in the view itself, with a value of `""`.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -97,7 +98,11 @@ class EditSetController @Inject() (
       FieldNames.startDateYear  -> text,
       FieldNames.endDateDay     -> text,
       FieldNames.endDateMonth   -> text,
-      FieldNames.endDateYear    -> text
+      FieldNames.endDateYear    -> text,
+      FieldNames.legalStatus -> text.verifying(
+        messagesApi("edit-set.record.error.legal-status")(Lang.apply("en")),
```
tna-erasmos, line 103: You could use `resolvedMessage`.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -97,7 +98,11 @@ class EditSetController @Inject() (
...
+        value => !value.equalsIgnoreCase("ref.0")
```
tna-erasmos, line 104: I believe the validation should be to assert that the value exists in the list of all legal statuses. Given that it's static for now, it can be done here.

tna-erasmos (commented):
A few suggested changes.

test/support/BaseSpec.scala
userjd: And I have also renamed this to `legalStatusRefereneData`

conf/messages
userjd: I don't follow this, do you mean group alpahbetically for errors?

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
userjd: Won't blank option be part of the list?

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
userjd: Would the validation in UI not suffice to check the value isn't empty or does the UI need to match the value wth actual the reference data?

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
tna-erasmos: Yes, I'd say that would be an improvement.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: By UI validation do you mean the form validation here?

In any case, there seem to be two types of validation: the first is where nothing was selected (a blank), and that can covered here.

The second is where the legal status associated with the particular record somehow doesn't match any of the possible legal statuses.

What should happen in that case? I don't recall you haven't a test scenario to cover that, but I might be wrong.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
rwalpole: I agree. This data should represent what will be received from the services API and that will not have anything with this text in. This should be created as a distinct `SelectItem` in the view and it should marked as disabled so that it cannot be selected.

rwalpole (changes requested):

conf/messages
userjd: The list is now sorted alphabetically which wasn't the case before.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
tna-erasmos: Good idea about the disabling it; I'll do the same for [this similar story](https://national-archives.atlassian.net/browse/PACT-607).

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: Just to know, there's a place for doing additional validation [here](https://github.com/nationalarchives/ctd-omega-editorial-frontend/blob/main/app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala#L195).

I do agree about the creators, but where we have the full list (as we will where we use a select box), we could do this validation after the normal form validation.

userjd:
@rwalpole added distinct `SelectItem` in the view and marked as disabled so that it cannot be selected. After this change however, the error message for legalStatus is replaced by the global `error.required`.

test/views/EditRecordViewSpec.scala
tna-erasmos: Confirmed.

test/support/CustomMatchers.scala
tna-erasmos: Confirmed.

test/support/BaseSpec.scala
tna-erasmos: Confirmed (both).

test/controllers/EditSetControllerSpec.scala

test/controllers/EditSetControllerSpec.scala

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
tna-erasmos: I believe we've agreed to postpone `disabled` for now.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: I'll assume that we're not doing this second validation: ensuring that it exists in the list.

Should be safe, as (as Rob said), they select from a list.

tna-erasmos (approved):
I've resolved all the comments I introduced.

rwalpole (approved):

PR #46, opened 2022-12-23
PACT572: add mock JSON data in the Web Application from 3 records to …

userjd:
…12 records

### DIFF
--- a/app/uk/gov/nationalarchives/omega/editorial/models/package.scala
+++ b/app/uk/gov/nationalarchives/omega/editorial/models/package.scala
@@ -48,7 +48,7 @@ package object editSetRecords {
   val editSetRecord2: EditSetRecord = Json
     .parse(
       """{
-        |  "ccr" : "COAL 80/80/1",
+        |  "ccr" : "COAL 80/80/2",
         |  "oci" : "COAL.2022.V2RJW.P",
         |  "scopeAndContent" : "Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths.",
         |  "coveringDates" : "1960",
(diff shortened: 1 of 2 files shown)


app/uk/gov/nationalarchives/omega/editorial/models/package.scala
```diff
@@ -83,10 +83,190 @@ package object editSetRecords {
     .validate[EditSetRecord]
     .get
 
+  val editSetRecord4: EditSetRecord = Json
+    .parse(
+      """{
+        |  "ccr" : "COAL 80/80/1",
```
rwalpole, line 89: Can we please increment the CCRs as well as the OCIs so that they appear as distinct records.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
```diff
@@ -123,6 +303,60 @@ package object editSets {
...
         |      "coveringDates" : "1960"
+        |    },
+        |    {
+        |      "ccr" : "COAL 80/80/3",
+        |      "oci" : "COAL.2022.V4RJW.P",
+        |      "scopeAndContent" : "Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths.",
+        |      "coveringDates" : "1960"
+        |    },
+        |    {
+        |      "ccr" : "COAL 80/80/3",
```
rwalpole, line 314: Can we please increment the CCRs as well as the OCIs so that they appear as distinct records. The CCRs and OCIs for both the records and the edit set should tie up so that we can see clearly they are separate records when testing and demo-ing.

rwalpole (changes requested):

PR #52, opened 2023-01-05
PACT-609 Add Related Material list to the Update Record Page in the Web Application

felixgb:
Here's a screenshot of the added list:
Screenshot 2023-01-05 at 13 47 56


app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -228,39 +232,52 @@ class EditSetController @Inject() (
...
-          }
-
-        case Some("discard") =>
-          Redirect(controllers.routes.EditSetController.discard(id, recordId))
-
-        case Some("calculateDates") => calculateDates(user, title)
-
-        // TODO Below added to handle error flow which could be a redirect to an error page pending configuration
-        case _ =>
+      val action = for {
```
felixgb, line 241: This was refactored because I needed to get an instance of the `EditSetRecord` by OCI - handling the case where a record was not found would have made this method too unwieldy in my opinion.

test/controllers/EditSetControllerSpec.scala
```diff
@@ -1149,6 +1162,7 @@ class EditSetControllerSpec extends BaseSpec {
 
           }
           "doesn't exist" in {
+            pending
```
felixgb, line 1165: I don't think the behaviour of this test is correct - I think that if a submit happens for a record that doesn't exist the response should be a bad request of some kind.

I'd like to get this checked and either delete this test or have confirmation of this behaviour before merging.

test/controllers/EditSetControllerSpec.scala
```diff
@@ -945,6 +946,18 @@ class EditSetControllerSpec extends BaseSpec {
               ExpectedSelectOption("1", "The National Archives, Kew", selected = true),
               ExpectedSelectOption("2", "British Museum, Department of Libraries and Archives"),
               ExpectedSelectOption("3", "British Library, National Sound Archive")
+            ),
+            relatedMaterial = Seq(
+              ExpectedRelatedMaterial(
+                linkHref = "#;",
```
felixgb, line 952: I assume that's what "At present the links will not de-referenceable" means - these links do not link anywhere at the minute. I assume in the future they can link to any kind of record or edit set.

app/uk/gov/nationalarchives/omega/editorial/models/RelatedMaterial.scala
```diff
@@ -0,0 +1,30 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package uk.gov.nationalarchives.omega.editorial.models
+
+import play.api.libs.json._
+
+case class RelatedMaterial(linkHref: String, linkText: String, description: Option[String] = None)
```
felixgb, line 26: Note that the link itself is not optional. This is my interpretation of the [ticket](https://national-archives.atlassian.net/browse/PACT-609) but does not match the prototype:

Screenshot 2023-01-06 at 12 56 10

(my prototype is pink so I can differentiate at a glance)

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -219,9 +235,29 @@
     },
     attributes = Map("rows" -> "5", "autofocus" -> "true")))
 
+
+  <label class="govuk-label">@messages("edit-set.record.related-material")</h3>
```
rwalpole, line 239: Rather than a label this should be a h3 as per the prototype, e.g. `<h3 class="govuk-heading-s">Related material</h3>`

rwalpole (changes requested):

test/views/EditRecordViewSpec.scala
```diff
@@ -60,9 +59,32 @@ class EditRecordViewSpec extends BaseSpec {
       LegalStatus("ref.4", "Welsh Public Record(s)")
     )
 
-  val emptyRecord: EditSetRecord = EditSetRecord.apply(
-    ccr = "",
-    oci = "",
+  val editSetRecord = EditSetRecord(
+    ccr = "COAL 80/80/1",
+    oci = "COAL.2022.V5RJW.P",
+    scopeAndContent = "",
```
userjd, line 65: for consistency the oci needs to match with the ccr above

test/controllers/EditSetControllerSpec.scala
rwalpole: Exactly right @felixgb

test/controllers/EditSetControllerSpec.scala
rwalpole: @felixgb In future if the record doesn't exist then this would create a new record but currently we are only looking to edit existing records, not create new ones, so a bad request seems reasonable for now.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
felixgb: Done - I used the label class before because the other labels are not bold but I imagine these will be changed to match the prototype

test/controllers/EditSetControllerSpec.scala
```diff
@@ -1148,48 +1161,6 @@ class EditSetControllerSpec extends BaseSpec {
             )
 
           }
-          "doesn't exist" in {
```
tna-erasmos, line 1151: Why was this scenario removed?

conf/messages
```diff
@@ -27,6 +27,7 @@ edit-set.record.edit.note=Note (optional)
 edit-set.record.edit.place-of-deposit=Held by
 edit-set.record.edit.scope-and-content=Scope and content
 edit-set.record.edit.start-date=Start date
+edit-set.record.related-material=Related material
```
tna-erasmos, line 30: Could you sort this list?

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -219,9 +239,31 @@
     },
     attributes = Map("rows" -> "5", "autofocus" -> "true")))
 
+
+  <h3 class="govuk-heading-s">@messages("edit-set.record.related-material")</h3>
+  <ul class="govuk-list govuk-list--bullet">
+    @{
+      staticRecord.relatedMaterial.map {
+
+        case e: RelatedMaterial.LinkAndDescription => {
```
tna-erasmos, line 248: You can remove the braces.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -219,9 +239,31 @@
...
+          relatedMaterialLinkAndDescription(e)
+        }
+
+        case e: RelatedMaterial.LinkOnly => {
```
tna-erasmos, line 252: You can remove the braces.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -219,9 +239,31 @@
...
+          relatedMaterialLink(e)
+        }
+
+        case e: RelatedMaterial.DescriptionOnly => {
```
tna-erasmos, line 256: You can remove the braces.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -38,6 +46,18 @@
     ))
 }
 
+@relatedMaterialLink(link: RelatedMaterial.LinkOnly) = {
```
tna-erasmos, line 49: I like this separation into small parts.

app/uk/gov/nationalarchives/omega/editorial/forms/EditSetRecordFormValues.scala
```diff
@@ -0,0 +1,81 @@
...
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package uk.gov.nationalarchives.omega.editorial.forms
+
+import uk.gov.nationalarchives.omega.editorial.models.EditSetRecord
+
+import play.api.data.{ Form, FormError }
```
tna-erasmos, line 26: Unused.

app/uk/gov/nationalarchives/omega/editorial/forms/EditSetRecordFormValues.scala
```diff
@@ -0,0 +1,81 @@
...
+import play.api.data.Forms.{ mapping, text }
```
tna-erasmos, line 27: Unused.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
```diff
@@ -42,11 +42,24 @@ package object editSetRecords {
...
+        |      "linkHref" : "#;",
+        |      "linkText" : "COAL 80/80/2",
+        |      "description" : "Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths."
+        |    }
+        |  ]
         |} """.stripMargin
     )
-    .validate[EditSetRecord]
-    .get
+    .as[EditSetRecord]
```
tna-erasmos, line 62: Why no 'validate' here and with the second, but in all other cases?

app/uk/gov/nationalarchives/omega/editorial/models/EditSet.scala
```diff
@@ -41,11 +41,12 @@ case class EditSetRecord(
...
   legalStatus: String,
   placeOfDeposit: String,
-  note: String
+  note: String,
+  relatedMaterial: Seq[RelatedMaterial] = Seq.empty
 )
 
 object EditSetRecord {
-  implicit val editSetRecordReads = Json.reads[EditSetRecord]
+  implicit val editSetRecordReads = Json.using[Json.WithDefaultValues].reads[EditSetRecord]
```
tna-erasmos, line 49: I believe we need this to handle those (json) records where the `relatedMaterial` is missing - but is that a valid assumption, as opposed to expecting an empty array? 

For example:

```
{
   "ccr":"COAL 80/80/12",
   "oci":"COAL.2022.V12RJW.P",
   "scopeAndContent":"Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths.",
   "coveringDates":"1960",
   "formerReferenceDepartment":"",
   "startDateDay":"1",
   "startDateMonth":"1",
   "startDateYear":"1960",
   "endDateDay":"31",
   "endDateMonth":"12",
   "endDateYear":"1960",
   "legalStatus":"ref.1",
   "placeOfDeposit":"1",
   "note":"",
   "relatedMaterial":[
      
   ]
}
```

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -446,4 +492,18 @@ class EditSetController @Inject() (
 
 object EditSetController {
   case class EditSetReorder(field: String, direction: String)
+
+  sealed abstract class SubmitAction
```
tna-erasmos, line 496: I like this.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -228,39 +232,52 @@ class EditSetController @Inject() (
...
+              corporateBodies.all,
+              formWithErrors
+            )
+          )
+
+        case Left(RecordNotFound(missingOci)) =>
+          BadRequest(s"Record with $missingOci not found")
+
+        case Left(InvalidAction(_)) =>
+          BadRequest("This action is not allowed")
```
tna-erasmos, line 276: Perhaps we should include the unknown action in the text? Or log it?

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -33,11 +33,14 @@ import uk.gov.nationalarchives.omega.editorial.models.{ DateRange, EditSetEntry,
...
 import uk.gov.nationalarchives.omega.editorial.services.CoveringDateError
 import uk.gov.nationalarchives.omega.editorial.support.DateParser
+import uk.gov.nationalarchives.omega.editorial.forms.EditSetRecordFormValues
+import uk.gov.nationalarchives.omega.editorial.forms.EditSetRecordFormValues._
 import uk.gov.nationalarchives.omega.editorial.views.html.{ editSet, editSetRecordEdit, editSetRecordEditDiscard, editSetRecordEditSave }
 
 import java.time.LocalDate
 import java.time.temporal.ChronoField.{ DAY_OF_MONTH, MONTH_OF_YEAR, YEAR }
 import javax.inject._
+import uk.gov.nationalarchives.omega.editorial.models.RelatedMaterial
```
tna-erasmos, line 43: Unused.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -96,12 +101,10 @@ class EditSetController @Inject() (
   private val orderDirectionAscending = "ascending"
   private val orderDirectionDescending = "descending"
 
-  type FormTransformer = Form[EditSetRecord] => Form[EditSetRecord]
+  type FormTransformer = Form[EditSetRecordFormValues] => Form[EditSetRecordFormValues]
 
-  private val editSetRecordForm: Form[EditSetRecord] = Form(
+  private val editSetRecordForm: Form[EditSetRecordFormValues] = Form(
```
tna-erasmos, line 106: Good - this only contains the editable fields.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -66,6 +70,7 @@ class EditSetController @Inject() (
     val note = "note"
     val oci = "oci"
     val placeOfDeposit = "placeOfDeposit"
+    val relatedMaterial = "relatedMaterial"
```
tna-erasmos, line 73: Unused.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
```diff
@@ -42,11 +42,24 @@ package object editSetRecords {
         |  "endDateYear" : "1962",
         |  "legalStatus": "ref.1",
         |  "placeOfDeposit" : "1",
-        |  "note": "A note about COAL.2022.V1RJW.P."
+        |  "note": "A note about COAL.2022.V1RJW.P.",
+        |  "relatedMaterial" : [
+        |    {
```
tna-erasmos, line 47: An empty object causes a RTE.

app/uk/gov/nationalarchives/omega/editorial/models/RelatedMaterial.scala
```diff
@@ -0,0 +1,51 @@
...
+
+package uk.gov.nationalarchives.omega.editorial.models
+
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+import play.api.libs.json.Reads._
+
+sealed abstract class RelatedMaterial
+
+object RelatedMaterial {
```
tna-erasmos, line 30: I like how you've done this. 

I would have expected a case class with two optional fields:  an object for the link with two strings, and a string for the description. 

But maybe this is better?

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -21,13 +21,21 @@
...
+@import uk.gov.nationalarchives.omega.editorial.forms.EditSetRecordFormValues
 @import uk.gov.hmrc.govukfrontend.views.html.components.implicits._
 
 @this(govukButton: GovukButton, govukInput: GovukInput, govukTextarea: GovukTextarea, govukFieldset: GovukFieldset, govukErrorSummary: GovukErrorSummary, govukSelect: GovukSelect)
-@(user: User, editSetName: String, title: String, legalStatusReferenceData: Seq[LegalStatus], corporateBodies: Seq[CorporateBody], editSetRecordForm: Form[EditSetRecord])(implicit messages: Messages, request: Request[AnyContent])
+@(
+  user: User,
+  editSetName: String,
+  title: String,
+  staticRecord: EditSetRecord,
```
tna-erasmos, line 34: Why not just `record`; why `static`?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -945,6 +946,18 @@ class EditSetControllerSpec extends BaseSpec {
               ExpectedSelectOption("1", "The National Archives, Kew", selected = true),
               ExpectedSelectOption("2", "British Museum, Department of Libraries and Archives"),
               ExpectedSelectOption("3", "British Library, National Sound Archive")
+            ),
+            relatedMaterial = Seq(
```
tna-erasmos, line 950: But there should be 3 items; the one without the link is missing.

tna-erasmos (changes requested):

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
felixgb: As in, an object like:
```
        |  "relatedMaterial" : [
        |    {
        |    },
...
```
?
The runtime exception is caused by the JSON object failing to parse as related material. I think that behaviour is expected?

app/uk/gov/nationalarchives/omega/editorial/models/RelatedMaterial.scala
felixgb: I thought about a single case class with two optional fields - one field for the link text and href and another for the description.  My issue would be the case where both of those fields were `None` - to me that is allowing an unintended state to be represented.

I think that parsing json would be just as complicated in that case as well. The only case where the default json parser could be used is if all three fields were present and optional on a single case class, but that would allow several unintended states to be represented.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: Good call.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
tna-erasmos: I think that we're using the `Label()` function in most cases.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
felixgb: I changed these json objects to use `.as[...]` instead of `.validate[...]` because `as` throws the json parse exception, whereas validate and then get gives you a non-descriptive exception with no parse error, so it's harder to see what failed.

The ones that still use `validate` got merged in from main, I will change them to use as.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
tna-erasmos: A safe assumption that it will have either a description or a link.

Of course, if the class consisted of two optional fields that this would pass.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
felixgb: I named this `staticRecord` to distinguish it from the form with "dynamic" values that change when this form is submitted, but I'm not sure about the wording, especially if other people find it unclear.

I'm happy to change it to be `record`, if you think that makes more sense?

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
felixgb: I did this because I was running into issues templating values inside a scala block inside a twirl block - breaking these out into named snippets makes it easier to work with.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
felixgb: So I can - done.

app/uk/gov/nationalarchives/omega/editorial/models/EditSet.scala
felixgb: I can't think of a case where you would want to differentiate between an empty collection and a property that doesn't exist for this specific case -

In general I use default parameters where I can to make refactoring easier. But if there's a case where the two cases have different meaning then I will change it.

test/controllers/EditSetControllerSpec.scala
tna-erasmos: The intent of that scenario is to show what happens when the start date doesn't exist - not the record (by OCI).

If it did refer to a record that didn't exist, that was accidental.

See the one below that for end date (line 1239).

conf/messages
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html

app/uk/gov/nationalarchives/omega/editorial/forms/EditSetRecordFormValues.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/forms/EditSetRecordFormValues.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/models/package.scala
tna-erasmos: Sounds like `as` is the way to go with all of them. Good.

Confirmed (all).

app/uk/gov/nationalarchives/omega/editorial/models/EditSet.scala
tna-erasmos: I'm not sure what the API contract will be.

My personal preference is to include all of the fields.

But that's not my call.

Let's call this resolved.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/models/RelatedMaterial.scala
tna-erasmos: That's a good point: that we're preventing an invalid state by modelling it this way.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
tna-erasmos: Up to you; but I'd suggest that we be consistent with `EditSetController`, either way.

test/views/EditRecordViewSpec.scala
tna-erasmos: It appears to match the record in `package` ... or perhaps I misunderstood.

test/controllers/EditSetControllerSpec.scala
tna-erasmos: Actually, you just need to change this line in the original scenario:

```
val result = submitWhileLoggedIn(1, "COAL.2022.V1RJW.P", values)
```

Incidentally, it was edit set ID `2` which works, but is misleading so best to change that too.

test/views/EditRecordViewSpec.scala
felixgb: It should match the record in the package file. I changed it to match after the first comment.

test/controllers/EditSetControllerSpec.scala
felixgb: Changed; I misunderstood the original test and though it was asserting there was a bad response for a missing OCI.

test/controllers/EditSetControllerSpec.scala
felixgb: Absolutely, I've changed the testing strategy now.

userjd (approved):

test/controllers/EditSetControllerSpec.scala
tna-erasmos: Confirmed.

tna-erasmos (approved):

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -40,6 +48,22 @@
     ))
 }
 
+@noRelatedMaterial = {
+  <li>None</li>
+}
+
+@relatedMaterialLink(link: RelatedMaterial.LinkOnly) = {
+  <li><a class="relatedMaterialLink" href="@link.linkHref">@link.linkText</a></li>
```
rwalpole, line 56: Can we please use all lower case attribute names within the HTML, so this should be `<a class="related-material-link"..>` as this is a W3C recommendation and it is how attributes are shown within the GOV.UK Design System.

rwalpole (changes requested):

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
tna-erasmos: That's good to know.

rwalpole (approved):

PR #54, opened 2023-01-05
PACT-646: adding deploy step to GitHub actions

rwalpole:


dist/scripts/ctd-omega-editorial-frontend.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=CTD Omega Editorial Frontend
+Documentation=https://github.com/nationalarchives/ctd-omega-editorial-frontend
+After=syslog.target
+
+[Service]
+Type=simple
+User=ctd-web
+Group=ctd-web
+Environment="JAVA_HOME=/usr/lib/jvm/jre-1.8.0
```
tna-erasmos, line 10: End double quote is missing.

.github/workflows/ci.yml
```diff
@@ -24,3 +24,31 @@ jobs:
...
+      - name: Configure AWS credentials for GitHub OIDC
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: arn:aws:iam::495195014394:role/IAMGitHubOIDCRole
+          aws-region: eu-west-2
+      - name: Copy package to S3 bucket
+        working-directory: target/universal
+        run: aws s3 cp *.zip s3://ctd-omega-frontend-deployment/
+      - name: Deploy package to EC2
+        run: aws deploy create-deployment --application-name CtdOmegaEditorialFrontend --deployment-group-name CtdOmegaEditorialFrontend-DepGrp --revision revision-type=S3 --s3-location bucket=ctd-omega-frontend-deployment,key=ctd-omega-editorial-frontend-0.1.0-SNAPSHOT.zip,bundleType=zip
```
tna-erasmos, line 54: The version (`ctd-omega-editorial-frontend-0.1.0-SNAPSHOT.zip`) is hardcoded. 

Is this temporary?

.github/workflows/ci.yml
```diff
@@ -24,3 +24,31 @@ jobs:
...
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up JDK 8
+        uses: actions/setup-java@v3
```
tna-erasmos, line 39: Why `@v3` here but `@v2` on line 19?

tna-erasmos (changes requested):
Mostly questions; the only required fix is to add the ending double quote.

dist/scripts/ctd-omega-editorial-frontend.service
```diff
@@ -0,0 +1,16 @@
...
+Environment="APPLICATION_SECRET=_mTFr@DIDmTUC9LFSFsn_j]iICDMj=DKMy/yuw/gy10RgsUiXuOK0beEP]SiCwMO"
```
felixgb, line 11: Are we leaving the secret in the repository until the application is deployed somewhere?
Maybe if/when this becomes an issue we could use `EnvironmentFile=` pointing to a file only readable by the ctd-web user.

dist/scripts/ctd-omega-editorial-frontend.service
rwalpole: Fixed.

.github/workflows/ci.yml
rwalpole: Yes, this isn't ideal. Either we have to skip version numbers and always use the same package name or we can try to attach the version number when the GitHub Actions run. I think this could be acheived by serailizing the BuildInfo to JSON and then parsing the JSON within the workflow. Not trivial though so I think we would need another ticket for that.

.github/workflows/ci.yml
rwalpole: No reason. Have now updated all actions to v3.

dist/scripts/ctd-omega-editorial-frontend.service
rwalpole: @felixgb That seems like a sensible approach. We should be able to use Terraform to create the file and put it on the server. I will check with @adamretter as I think he has done something like this previously.

dist/scripts/ctd-omega-editorial-frontend.service
tna-erasmos: Confirmed.

.github/workflows/ci.yml
tna-erasmos: Agreed; could we introduce a placeholder ticket for that work (just a minimal description)?

.github/workflows/ci.yml
tna-erasmos: Confirmed.

tna-erasmos (approved):

felixgb (approved):

PR #61, opened 2023-01-16
PACT-575 Investigate existing options for ordering of columns in lists of results in the Web Application

felixgb:
Table on page load, note that none of the table headers indicate a sort:
Screenshot 2023-01-18 at 09 52 28
Sort by Name, note the header arrows:
Screenshot 2023-01-18 at 09 53 01
Sorting is stable. First I sorted by "Elevation", descending. Row with "Test2" appears above "Aconcagua" - then I sorted by "Continent". "Test2" still appears above "Aconcagua".
Screenshot 2023-01-18 at 09 53 34

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -26,3 +26,70 @@
 .govuk-width-container {
     max-width: 70em;
 }
+
+[aria-sort] button,
+[aria-sort] button:hover {
+  background-color: transparent;
+  border-width: 0;
+  -webkit-box-shadow: 0 0 0 0;
+  -moz-box-shadow: 0 0 0 0;
+  box-shadow: 0 0 0 0;
+  color: #005ea5;
+  cursor: pointer;
+  font-family: inherit;
+  font-size: inherit;
+  font-weight: inherit;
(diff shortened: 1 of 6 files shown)


felixgb:
TODO:
[ ] numeric cell types as well string types, these should right align: https://design-system.service.gov.uk/components/table/

rwalpole (commented):
This seems to be exactly what we want and it's clearly not reliant on JavaScript. Where did it come from? Is there any user/accessibilty testing information?

felixgb:
The solution is partially inspired from this sortable table example from W3C:
https://www.w3.org/WAI/ARIA/apg/example-index/table/sortable-table
However, that solutions uses javascript.

I've included the `aria-sort`attribute on the sortable column, however the W3C table has a few more features:
- Hover css classes to indicate the headers are clickable and to highligh which headers are selectable only using keyboard
- A hidden span caption element for screenreaders:
```
  <table class="sortable">
    <caption>
      Students currently enrolled in WAI-ARIA 101
      <span class="sr-only">
        , column headers with buttons are sortable.
      </span>
    </caption>
    ....
```

PR #65, opened 2023-01-18
[PACT-657] Introduced integration tests, namely for the login page.

tna-erasmos:


it/BaseISpec.scala
```diff
@@ -0,0 +1,37 @@
+import org.jsoup.Jsoup
```
felixgb, line 1: Should all of these classes be inside a package?

I thought that classes defined with no package cannot be imported into another class, e.g.
```
packge random

class  RandomISpec extends BaseISpec {
...
}
```
would not compile because `BaseISpec` could not be found.

tna-erasmos: I can add a package.

But are we expecting any packages?

I had imagined only having one `ISpec` for each screen, in that folder.

felixgb: OK, I don't mind keeping it all at top level for now. Maybe in the future we'll have separate sets of integration tests, and we can split it out then. But no problem for now.

it/LoginISpec.scala
```diff
@@ -0,0 +1,266 @@
...
+
+  private def getEditSetPage(sessionCookie: WSCookie): WSResponse =
+    await(
+      wsUrl("/edit-set/1")
+        .withFollowRedirects(false)
+        .addCookies(sessionCookie)
+        .get()
+    )
+
+  private def getLoginPage: WSResponse =
```
felixgb, line 189: I  would add empty parens "()" after this method, to make it clear that it's side-effecting, see https://docs.scala-lang.org/style/method-invocation.html

runBeforePushing.sh
```diff
@@ -0,0 +1,2 @@
+sbt clean test IntegrationTest/test scalafmtAll
```
felixgb, line 1: Did you mean to add this file?

it/LoginISpec.scala
tna-erasmos: Sure. But what's the side-effect in this case?

felixgb: The HTTP client is making a GET request to the server - calling it two times may result in different values, for example if the HTTP server is not responding.

tna-erasmos: Ah OK - in a functional sense, yes.

I took getting the data from the server as the main effect.

But I'll change it.

runBeforePushing.sh
tna-erasmos: Certainly. It's referenced in the README, as well.

it/LoginISpec.scala
tna-erasmos: Changed.

runBeforePushing.sh
rwalpole: @tna-erasmos can you also please add `IntegrationTest /scalafmt` to the fmt alias in `.sbtrc`

tna-erasmos: Will do.

tna-erasmos: I upgraded the `scalafmt` dep from `2.4.3` to `2.4.4`, as this will automatically include the integration tests, for both `scalafmtAll` and `scalafmtCheckAll`.

I also simplified the sbt aliases.

PR #70, opened 2023-01-24
PACT-672: Remove LoginControllerSpec and LoginViewSpec

userjd:


tna-erasmos (approved):
Jaishree, Felix and I went through this, discussing a few cases which we might have want to keep, but agreed that they were sufficiently covered by the integration tests.

PR #75, opened 2023-01-30
[PACT-711] Added scoverage.

tna-erasmos:


rwalpole (approved):

rwalpole:
The report makes interesting reading! Once we are getting our data from the message fabric we should introduce a minimum coverage threshold. Personally I think we should aim for a minimum total coverage of around 85% - which shouldn't be too difficult to achieve - but we can discuss as a team.

felixgb (approved):

tna-erasmos:
The sore thumb in that report is `ReferenceDataService`, of course. It's not used in testing per se, but rather `TestReferenceDataService` which inherits from it.

Perhaps with the integration with the API, we'll resolve that.

PR #76, opened 2023-01-31
PACT-696 Display the record type next to the PAC-ID

userjd:


app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetRecordController.scala
```diff
@@ -309,11 +309,13 @@ class EditSetRecordController @Inject() (
     form: Form[EditSetRecordFormValues]
   )(implicit request: Request[AnyContent]): HtmlFormat.Appendable = {
     val title = resolvedMessage(MessageKeys.title)
+    val recordType = if (editSetRecord.oci.endsWith(".P")) resolvedMessage(MessageKeys.recordTypePhysical) else ""
```
felixgb, line 312: Do you think we should put this expression inside the view? It's just formatting, after all, and we're already passing the record in.

The advantage would be to cut down on the number of arguments we are passing into the view, which can get a bit unwieldy.

test/controllers/EditSetRecordControllerSpec.scala
```diff
@@ -407,10 +411,20 @@ class EditSetRecordControllerSpec extends BaseSpec {
             custodialHistory = "",
             relatedMaterial = Seq.empty,
             separatedMaterial = Seq.empty
-          )
+          ).copy()
```
felixgb, line 414: There's an extra copy in here?

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetRecordController.scala
userjd: @felixgb Yes can do, but the intention was to generated the type from the controller as I understand there could be possible changes needed if there are other types and it would bring the logic into the view

tna-erasmos: I agree with Felix.

And it's probably not too early to introduce a `recordType` field to `EditSetRecord`; for now, it can be a derived, optional field in that it's value depends on the OCI suffix.

test/support/CommonMatchers.scala
```diff
@@ -101,6 +101,13 @@ object CommonMatchers {
       actualValue = document.select("h1.govuk-heading-l").text()
     )
 
+  def haveHeadingPacId(expectedValue: String): Matcher[Document] = (document: Document) =>
```
tna-erasmos, line 104: Perhaps `havePacIDHeading` or `haveHeadingForPacID`?

test/controllers/EditSetRecordControllerSpec.scala
```diff
@@ -2601,6 +2624,7 @@ class EditSetRecordControllerSpec extends BaseSpec {
     ExpectedEditRecordPage(
       title = "Edit record",
       heading = s"TNA reference: ${editSetRecord.ccr}",
+      headingPacId = s"PAC-ID: ${editSetRecord.oci} Physical Record",
```
tna-erasmos, line 2627: This wouldn't work for a record without a type (e.g. `COAL.2022.V2RJW`).

test/controllers/EditSetRecordControllerSpec.scala
```diff
@@ -410,7 +414,25 @@ class EditSetRecordControllerSpec extends BaseSpec {
...
       }
-
+      "all data is valid with no record type suffix" in {
+        val oci = "COAL.2022.V2RJW"
+        val getRecordResult = getRecordForEditingWhileLoggedIn(1, oci)
+        assertPageAsExpected(
+          asDocument(getRecordResult),
+          generateExpectedEditRecordPageFromRecord(oci).copy(
+            headingPacId = s"PAC-ID: $oci",
+            optionsForCreators = Seq(
```
tna-erasmos, line 424: We probably don't need to include 'optionsForCreators'; the important part to assert is the `headingPacId`, which you already have.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
```diff
@@ -443,7 +444,7 @@ <h3 class="govuk-heading-s">@messages("edit-set.record.edit.separated-material")
         @govukErrorSummary(ErrorSummary().withFormErrorsAsText(editSetRecordForm))
     }
     <h1 class="govuk-heading-l">@messages("edit-set.record.edit.heading", record.ccr)</h1>
-    <h2 class="govuk-heading-m">@messages("edit-set.record.edit.id", record.oci)</h2>
+    <h2 class="govuk-heading-m">@messages("edit-set.record.edit.id", record.oci) @recordType</h2>
```
tna-erasmos, line 447: Couldn't we send the record type as the second arg to the message?

tna-erasmos (commented):
Only suggestions.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetRecordController.scala
tna-erasmos: For example, in `EditSetRecord`:

```
  sealed trait RecordType
  
  case object Physical extends RecordType
```

```
case class EditSetRecord(
  ccr: String,
  oci: String,
  scopeAndContent: String,
  coveringDates: String,
  formerReferenceDepartment: String,
  formerReferencePro: String,
  startDateDay: String,
  startDateMonth: String,
  startDateYear: String,
  endDateDay: String,
  endDateMonth: String,
  endDateYear: String,
  legalStatusID: String,
  placeOfDepositID: String,
  note: String,
  background: String,
  custodialHistory: String,
  relatedMaterial: Seq[RelatedMaterial] = Seq.empty,
  separatedMaterial: Seq[SeparatedMaterial] = Seq.empty,
  creatorIDs: Seq[String] = Seq.empty
) {

  val recordType: Option[RecordType] = if (oci.endsWith(".P")) Option(Physical) else None

}
```

test/support/CommonMatchers.scala
tna-erasmos: Or maybe even `haveSubheading`?

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetRecordController.scala
felixgb: That solution looks good, both to keep the logic out of the view and out of the controller actions.

userjd: Done

test/support/CommonMatchers.scala
userjd: Done

test/controllers/EditSetRecordControllerSpec.scala
userjd: For `COAL.2022.V2RJW` it works by checking for this value `subHeading = s"PAC-ID: $oci"`instead as seen in test _"all data is valid with no record type suffix"_

test/controllers/EditSetRecordControllerSpec.scala
userjd: That's true but the test expects the default `Select creator` as the selected option or it fails with _We expected 0 creator selections, but there were actually 1._

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
userjd: Done

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetRecordController.scala
tna-erasmos: Confirmed.

test/controllers/EditSetRecordControllerSpec.scala
tna-erasmos: Yes, I see that it works. But by coincidence.

However, if you used that same record and asserting something different, it would fail.

If you wanted to fix it, you could dynamically determine the correct type.

test/controllers/EditSetRecordControllerSpec.scala
tna-erasmos: Ah - you found a bug.

Starting on line 2639, this will fix it:

```
      optionsForCreators = {
        val recognisedCreatorIds = editSetRecord.creatorIDs.filter(creatorId => allCreators.exists(_.id == creatorId))
        val correctedCreatorIds = if (recognisedCreatorIds.nonEmpty) recognisedCreatorIds else Seq("")
        correctedCreatorIds
        .map(creatorId =>
          Seq(
            ExpectedSelectOption("", "Select creator", disabled = true),
            ExpectedSelectOption("48N", "Baden-Powell, Lady Olave St Clair (b.1889 - d.1977)"),
            ExpectedSelectOption("46F", "Fawkes, Guy (b.1570 - d.1606)", selected = true),
            ExpectedSelectOption("92W", "Joint Milk Quality Committee (1948 - 1948)"),
            ExpectedSelectOption("8R6", "Queen Anne's Bounty")
          ).map(expectedSelectedOption =>
            expectedSelectedOption.copy(selected = expectedSelectedOption.value == creatorId)
          )
        )},
```

Or adapt as you wish.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html
tna-erasmos: Almost there - you just need to resolve that into a message.

As it is, it comes out as `PhysicalRecord` instead of `Physical Record`.

test/support/CommonMatchers.scala
```diff
@@ -101,6 +101,13 @@ object CommonMatchers {
       actualValue = document.select("h1.govuk-heading-l").text()
     )
 
+  def haveSubHeading(expectedValue: String): Matcher[Document] = (document: Document) =>
+    singleValueMatcher(
+      label = "a heading",
```
tna-erasmos, line 106: Good. But should you just change the label to `a subheading`?

test/support/CommonMatchers.scala
tna-erasmos: Confirmed.

test/controllers/EditSetRecordControllerSpec.scala
userjd: Done

test/controllers/EditSetRecordControllerSpec.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/views/editSetRecordEdit.scala.html

test/support/CommonMatchers.scala
tna-erasmos: Confirmed.

tna-erasmos (approved):
LGTM

PR #77, opened 2023-02-02
[PACT-712] Spike: Embedded the JMS EchoServer.

tna-erasmos:


tna-erasmos:
Please see this comment about [the failing jobs](https://national-archives.atlassian.net/browse/PACT-712?focusedCommentId=40033).

app/modules/StartupModule.scala
```diff
@@ -0,0 +1,56 @@
...
+import com.google.inject.AbstractModule
+import uk.gov.nationalarchives.omega.editorial.EchoServerBootstrap
+
+import scala.annotation.unused
+
+@unused // Found by Play, via application.conf (play.modules.enabled)
+class StartupModule extends AbstractModule {
+
+  override def configure(): Unit =
+    bind(classOf[EchoServerBootstrap]).asEagerSingleton()
```
felixgb, line 54: Does this stop the server when the application is stopped? i.e. can you run the application in the sbt console, and have the echo server stop when the application is stopped? Or is it shutdown when the JVM is stopped?

tna-erasmos: When you stop `sbt`, it appears that the echo server is stopped, in that it no longer echoes; but how graciously, I don't know; if you have a look at  `IOApp.main` it does handle the JVM shutdown; but of course, it's ultimately up to the resources (like the [JMS Client](https://github.com/nationalarchives/ctd-omega-editorial-frontend/blob/3c0dbfc899ac2111e18a8ce83d5b2fba97d73f1e/app/uk/gov/nationalarchives/omega/editorial/services/EchoServer.scala#L45-L450) how graceful it is.

P.S. This is my first exposure to Cats Effect / IO, so there's probably a tidier way of embedding an IO application.

tna-erasmos:
You can find a demo video, [here](https://drive.google.com/file/d/12Wo3m7WieVX8YSz2ejOHfKYHcvTXOUB9/view).

rwalpole:
Thanks for the demo video @tna-erasmos. As soon as I saw it I realised I had forgot the old "Play does nothing until you kick it" rule. Now working ok.

PR #79, opened 2023-02-06
[PACT-698] Added integration test for page with edit set records.

tna-erasmos:


felixgb (approved):
Looks good -- I can't find much to comment about.
2 general points:
* I assume we are going to delete the EditSetSpec in another ticket?
* If you _wanted_ you could add:
```
Test / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oD")
IntegrationTest / testOptions += Tests.Argument(TestFrameworks.ScalaTest, "-oD")
```
to the build.sbt, that would show the elapsed time for each test, which I think is somewhat useful.

tna-erasmos:
@felixgb , if you mean removing `EditSetControllerSpec`, then I would have thought so ([as we did with the login page](https://national-archives.atlassian.net/browse/PACT-657)), but @rwalpole has some [concerns](https://national-archives.atlassian.net/browse/PACT-698?focusedCommentId=40248) about that.

Thanks for that tip about timing; I'll try it now.

it/EditSetISpec.scala
```diff
@@ -0,0 +1,1099 @@
+import controllers.EditSetControllerSpec.ExpectedEditSetSummaryRow
+import play.api.http.Status.OK
+import play.api.libs.ws.{ DefaultWSCookie, WSCookie, WSResponse }
+import play.api.test.Helpers.{ await, defaultAwaitTimeout }
+import support.CommonMatchers._
+
+class EditSetISpec extends BaseISpec {
+
+  "when viewed, for a specified edit set ID" should {
```
rwalpole, line 9: Could we use `must` rather than `should` please.

tna-erasmos: Sure. I'll change them.

tna-erasmos: I'll change them in `LoginISpec`, as well.

tna-erasmos: Changed.

rwalpole: Thanks

rwalpole (approved):

PR #85, opened 2023-02-21
PACT-735 Added response header to prevent browser caching of pages

userjd:


app/uk/gov/nationalarchives/omega/editorial/support/filter/ResponseHeaderFilter.scala
```diff
@@ -0,0 +1,37 @@
...
+package uk.gov.nationalarchives.omega.editorial.support.filter
+
+import akka.stream.Materializer
+import play.api.Logging
+import play.api.mvc._
+
+import javax.inject.Inject
+import scala.concurrent.{ ExecutionContext, Future }
+
+class ResponseHeaderFilter @Inject() (implicit val mat: Materializer, ec: ExecutionContext)
```
rwalpole, line 31: I think we should just call this `CacheControlFilter` as that is all it does currently. The actually filtering is done on the request header and the result is the response header. We are not filtering the response header as such.

rwalpole (changes requested):

app/uk/gov/nationalarchives/omega/editorial/support/filter/ResponseHeaderFilter.scala
userjd: Done

app/uk/gov/nationalarchives/omega/editorial/support/filter/CustomFilter.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package uk.gov.nationalarchives.omega.editorial.support.filter
+
+import play.api.http.DefaultHttpFilters
+import play.filters.cors.CORSFilter
+import play.filters.csrf.CSRFFilter
+
+import javax.inject.Inject
+
+class CustomFilter @Inject() (
```
rwalpole, line 30: Can we call this `CustomFilters` with an `s` as it is really a collection of filters.

rwalpole (changes requested):

app/uk/gov/nationalarchives/omega/editorial/support/filter/CustomFilter.scala
userjd: Done

rwalpole (approved):

PR #86, opened 2023-02-21
[PACT-724] Added the JMS Request / Reply Client.

tna-erasmos:


rwalpole (approved):

userjd:
Integration tests are working with success scenarios. Just a few minor suggestions:

- adding inline comments (from the original `jms4s-request-repl`y) would be helpful for understanding the code and usage of cats effect better.  
- adding error scenarios in Integration tests for eg. incorrect queue or service id  unless there is never the possibility for this

tna-erasmos:
> adding inline comments (from the original jms4s-request-reply) would be helpful for understanding the code and usage of cats effect better.

I didn't find they added much, but am open to restoring them. @rwalpole ?

> adding error scenarios in Integration tests for eg. incorrect queue or service id unless there is never the possibility for this

Agreed - I could have added a number of such scenarios; but I understood the task as relocating the code from the original repo with minor changes.

PR #87, opened 2023-02-21
[PACT-762] Now starting ElasticMQ via Testcontainers.

tna-erasmos:


felixgb (approved):
I think this does indeed start the elastic-mq container for the JMS request reply integration test. However, I cannot confirm that it works because the tests hang forever on my machine. Occasionally the first spec runs:
```
[info] SQS Client
[info] - send a message and handle the reply (1 second, 516 milliseconds)
```
I've confirmed this is a problem on main, not just on the branch. I'm working on finding the cause of this but in the meantime I think this PR demonstrates that testcontainers can be used.

felixgb:
> I think this does indeed start the elastic-mq container for the JMS request reply integration test. However, I cannot confirm that it works because the tests hang forever on my machine. Occasionally the first spec runs:
> 
> ```
> [info] SQS Client
> [info] - send a message and handle the reply (1 second, 516 milliseconds)
> ```
> 
> I've confirmed this is a problem on main, not just on this branch. I'm working on finding the cause of this but in the meantime I think this PR demonstrates that testcontainers can be used.

This passes now for me-- after restarting the service several times.

tna-erasmos:
> This passes now for me-- after restarting the service several times.

@felixgb - which service are you restarting?

felixgb:
> @felixgb - which service are you restarting?

Sorry, I should have been more clear. I'm killing SBT, removing all docker containers (not that this should make a difference, as no data is preserved on any mounted volume) then starting it again. It seems to be pretty consistent for me now, though.

rwalpole (approved):

PR #90, opened 2023-03-01
PACT-525 Create ApiConnector

felixgb:


test/support/BaseSpec.scala
```diff
@@ -51,10 +55,21 @@ class BaseSpec extends PlaySpec with GuiceOneAppPerSuite with Injecting with Bef
   val validSessionToken: String = Session.generateToken("1234")
   val invalidSessionToken: String = Session.generateToken("invalid-user")
 
+  private lazy val testTimeProvider: TimeProvider = () => LocalDateTime.of(2023, Month.FEBRUARY, 28, 1, 1, 1)
+  private lazy val testConfig: Config = Config(
+    broker = HostBrokerEndpoint("not.a.real.host", 0),
```
felixgb, line 60: I wanted to make it explicit that this isn't meant to connect to anything, and make it an easy string to search for (as opposed to "localhost")

it/ApiConnectorISpec.scala
```diff
@@ -0,0 +1,53 @@
...
+        editSetResponse2 <- apiConnector.getEditSet(idOfExistingEditSet)
+      } yield (editSetResponse1, editSetResponse2)
+
+      val (editSetResponse1, editSetResponse2) = action.unsafeRunSync()
+
+      editSetResponse1 mustBe editSets.editSet1
+      editSetResponse2 mustBe editSets.editSet1
+    }
+
+    s"get the edit set 100 times for id: $idOfExistingEditSet" in {
```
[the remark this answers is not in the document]
tna-erasmos: Yes, I think it was reasonable to assume that we'd take a performance hit from treating sync operations as async.

I'd say that we remove this test.

If we are concerned about performance, perhaps we could introduce Gatling.

app/uk/gov/nationalarchives/omega/editorial/modules/StartupModule.scala
```diff
@@ -21,12 +21,29 @@
...
 
 class StartupModule extends AbstractModule {
 
   override def configure(): Unit =
     bind(classOf[StubServerBootstrap]).asEagerSingleton()
 
+  @Provides
+  @Singleton
+  def configProvider: Config =
+    ConfigSource.default.loadOrThrow[Config]
```
tna-erasmos, line 42: The lib seems like a good addition - makes for elegant code.

app/uk/gov/nationalarchives/omega/editorial/modules/StartupModule.scala
```diff
@@ -21,12 +21,29 @@
...
+
+  @Provides
+  @Singleton
+  def defaultTimeProvider: TimeProvider =
```
tna-erasmos, line 46: Useful to have for testing.

it/ApiConnectorISpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  lazy val apiConnector: ApiConnector = app.injector.instanceOf[ApiConnector]
+
+  "when a request for an edit set is made" must {
+    s"get an edit set for id: $idOfExistingEditSet" in {
+      val editSetResponse = apiConnector.getEditSet(idOfExistingEditSet).unsafeRunSync()
+
+      editSetResponse mustBe editSets.editSet1
+    }
+
+    s"get the edit set twice for id: $idOfExistingEditSet" in {
```
tna-erasmos, line 20: I wonder if we need this.

In order to verify the request / reply client? 

If so, we already do that.

Also, as we're returning the same edit set, there's no way to catch out an error by which one request gets the response meant for the other - which I think is the intent of such a test: to confirm the pairing.

test/support/TestApiConnector.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import cats.effect.IO
+
+import uk.gov.nationalarchives.omega.editorial.connectors.ApiConnector
+import uk.gov.nationalarchives.omega.editorial.models.EditSet
+import uk.gov.nationalarchives.omega.editorial.editSets
+
+object TestApiConnector extends ApiConnector(null, null, null) {
+
+  override def getEditSet(id: String): IO[EditSet] =
```
tna-erasmos, line 32: Could we get the requested Edit Set from the `package`?

Otherwise, we're limited to just the one, in the "unit" tests.

conf/application.conf
```diff
@@ -3,4 +3,19 @@ play.i18n.langs=["en"]
 play.http.secret.key="changeme"
 play.http.secret.key=${?APPLICATION_SECRET}
 play.modules.enabled += "uk.gov.nationalarchives.omega.editorial.modules.StartupModule"
-play.http.filters=uk.gov.nationalarchives.omega.editorial.support.filter.CustomFilters
\ No newline at end of file
+play.http.filters=uk.gov.nationalarchives.omega.editorial.support.filter.CustomFilters
+
+# JMS client takes appox 20 seconds to shut down - we increase the coordinated
```
tna-erasmos, line 8: Good and useful comment.

However - there is a typo: "appox".

app/uk/gov/nationalarchives/omega/editorial/services/EditSetService.scala
```diff
@@ -21,20 +21,22 @@
...
+import scala.concurrent.Future
 
 @Singleton
-class EditSetService {
+class EditSetService @Inject() (apiConnector: ApiConnector) {
 
   private val logger: Logger = Logger(this.getClass)
 
-  def getEditSet(id: String): EditSet = {
+  def getEditSet(id: String): Future[EditSet] = {
```
tna-erasmos, line 37: Given the context, could this merely name this `get`?

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
```diff
@@ -63,28 +65,28 @@ class EditSetController @Inject() (
...
+      val title = resolveMessage("edit-set.title", editSetPage.pageNumber, editSetPage.totalPages)
+      val heading: String = resolveMessage(
+        "edit-set.heading",
+        currentEditSet.name,
+        editSetPage.numberOfFirstEntry.toString,
+        editSetPage.numberOfLastEntry.toString,
+        editSetPage.totalNumberOfEntries.toString
+      )
+      Ok(editSet(user, title, heading, editSetPage, editSetEntryRowOrder))
+    }
```
tna-erasmos, line 89: It's outside the scope, but I wonder if you could remove the unknown members in the companion object?

They are: `EditSetReorder`, `orderDirectionAscending` and `orderDirectionDescending`.

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
```diff
@@ -0,0 +1,94 @@
...
+  private def parse[A : Reads](messageText: String): IO[A] =
+    IO.fromOption(
+      Json.parse(messageText).validate[A].asOpt
+    )(CannotParseEditSetResponse(messageText))
+
+}
+
+object ApiConnector {
+
+  case class CannotParseEditSetResponse(response: String) extends Exception(
```
tna-erasmos, line 88: Can be private.

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
```diff
@@ -0,0 +1,94 @@
...
+class ApiConnector @Inject() (
+  config: Config,
+  timeProvider: TimeProvider,
+  lifecycle: ApplicationLifecycle
+) {
+  import ApiConnector._
+
+  private implicit val logger: SelfAwareStructuredLogger[IO] = Slf4jLogger.getLogger[IO]
+
+  private val getEditSetSID = "OSGEES001"
```
tna-erasmos, line 46: I was going to say that this was duplicated from `ResponseBuilder` - but that one belongs to the `StubServer`, which is only used for testing. 

Given this, I wonder if we should move the latter two into the test source tree?

test/support/BaseSpec.scala
tna-erasmos: Yes. And BaseSpec will soon be slimmed down significantly as part of making these actual unit tests.

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
tna-erasmos: Out of curiosity, how would you use the stub server outside of automated tests?

tna-erasmos:
I forgot to [share the results of my QA](https://national-archives.atlassian.net/browse/PACT-525?focusedCommentId=42349). Works well enough.

it/ApiConnectorISpec.scala
felixgb: I just thought that it would demonstrate that once the request-reply client can handle 2 messages in succession, i.e. one is pulled off the queue before the other.

But I think you are right, that is a test for the request-reply client spec. I think I'll delete this one.

test/support/TestApiConnector.scala
felixgb: Sorry, I'm not sure if I understand. We currently only have one edit set in the package object. Do you think we should change the way that edit sets are represented in the package object so that we can have more than one of them, indexed by this `id`?

conf/application.conf
felixgb: Oof, thanks, fixed.

app/uk/gov/nationalarchives/omega/editorial/services/EditSetService.scala
felixgb: I think that makes sense, yes.

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
felixgb: Hm, I'm not sure about making case classes like this one private.
I would hope that another class that uses this one would catch this exception by pattern matching, at some point. Otherwise there's very little point having this custom exception.

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
felixgb: The stub server is enabled here, in the default configuration: https://github.com/nationalarchives/ctd-omega-editorial-frontend/blob/main/conf/application.conf#L5

it/ApiConnectorISpec.scala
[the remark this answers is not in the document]
felixgb: Happy to remove the test after another reviewer sees this comment.

it/ApiConnectorISpec.scala
tna-erasmos: Confirmed.

test/support/TestApiConnector.scala
tna-erasmos: Sorry - I confused myself: I was thinking of Edit Set Records, of which there are several.

conf/application.conf
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/services/EditSetService.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/controllers/EditSetController.scala
tna-erasmos: Confirmed.

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
tna-erasmos: OK

app/uk/gov/nationalarchives/omega/editorial/connectors/ApiConnector.scala
tna-erasmos: Eventually, I see this running against the real backend (aka `ctd-omega-services`).

So I see the stub service as a permanent addition for automated testing, but only a stop gap for manual running.

As such: yes, it should stay where it is.

it/ApiConnectorISpec.scala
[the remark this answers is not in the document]
tna-erasmos: In that case, I won't resolve it yet.

tna-erasmos (approved):
LGTM. Will be a good base for [PACT-526](https://national-archives.atlassian.net/browse/PACT-526).

app/uk/gov/nationalarchives/omega/editorial/services/EditSetService.scala
```diff
@@ -21,20 +21,22 @@
...
-class EditSetService {
+class EditSetService @Inject() (apiConnector: ApiConnector) {
 
   private val logger: Logger = Logger(this.getClass)
 
-  def getEditSet(id: String): EditSet = {
+  def get(id: String): Future[EditSet] = {
     logger.info(s"The edit set id is $id ")
-    editSets.getEditSet()
+    apiConnector.getEditSet(id).unsafeToFuture()
```
rwalpole, line 39: Use of `Future` should be restricted to the controller. This function should return `IO[EditSet]` and the `unsafeToFuture()` should be called at the end of controller method.

it/ApiConnectorISpec.scala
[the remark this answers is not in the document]
rwalpole: We are not concerned about peformance at the moment - but we will keep an eye on it.

app/uk/gov/nationalarchives/omega/editorial/services/EditSetService.scala
felixgb: That's moved now.

rwalpole (approved):

PR #92, opened 2023-03-03
PACT-733 Replace integration tests in EditSetControllerSpec with unit…

userjd:
… tests

### DIFF
--- a/test/controllers/EditSetControllerSpec.scala
+++ b/test/controllers/EditSetControllerSpec.scala
@@ -23,6 +23,8 @@ package controllers
 
 import cats.effect.IO
 import org.mockito.ArgumentMatchers.anyString
+
+import scala.concurrent.ExecutionContext.Implicits.global
 import play.api.i18n.Messages
 import play.api.mvc._
 import play.api.test._
--- a/test/controllers/EditSetControllerSpec.scala
+++ b/test/controllers/EditSetControllerSpec.scala
@@ -31,7 +31,7 @@ import play.api.test.Helpers._
 import scala.concurrent.ExecutionContext.Implicits.global
 import uk.gov.nationalarchives.omega.editorial.controllers.{ EditSetController, SessionKeys }
 import play.twirl.api.Html
-import support.BasePlaySpec
+import support.BaseControllerSpec
 import uk.gov.nationalarchives.omega.editorial.editSets.getEditSet
 import uk.gov.nationalarchives.omega.editorial.models.User
 import uk.gov.nationalarchives.omega.editorial.services.EditSetPagination.EditSetPage
@@ -43,7 +43,7 @@ import uk.gov.nationalarchives.omega.editorial.views.html.editSet
   *
   * For more information, see https://www.playframework.com/documentation/latest/ScalaTestingWithScalaTest
   */
-class EditSetControllerSpec extends BasePlaySpec {
+class EditSetControllerSpec extends BaseControllerSpec {
   override def beforeEach(): Unit = {
     super.beforeEach()
     restoreOriginalRecords()
--- a/test/controllers/EditSetControllerSpec.scala
+++ b/test/controllers/EditSetControllerSpec.scala
@@ -30,7 +30,6 @@ import play.api.mvc._
 import play.api.test._
 import play.api.test.Helpers._
 
-import scala.concurrent.ExecutionContext.Implicits.global
 import uk.gov.nationalarchives.omega.editorial.controllers.{ EditSetController, SessionKeys }
 import play.twirl.api.Html
 import support.BaseControllerSpec
--- a/test/controllers/EditSetControllerSpec.scala
+++ b/test/controllers/EditSetControllerSpec.scala
@@ -33,6 +33,8 @@ import uk.gov.nationalarchives.omega.editorial.controllers.{ EditSetController, 
 import uk.gov.nationalarchives.omega.editorial.editSetRecords.restoreOriginalRecords
 import uk.gov.nationalarchives.omega.editorial.editSets.getEditSet
 import uk.gov.nationalarchives.omega.editorial.models.User
+import uk.gov.nationalarchives.omega.editorial.services.Direction.{ Ascending, Descending }
+import uk.gov.nationalarchives.omega.editorial.services.EditSetEntryRowOrder.{ CCROrder, CoveringDatesOrder, ScopeAndContentOrder }
 import uk.gov.nationalarchives.omega.editorial.services.EditSetPagination.EditSetPage
 import uk.gov.nationalarchives.omega.editorial.services.{ EditSetEntryRowOrder, EditSetService }
 import uk.gov.nationalarchives.omega.editorial.views.html.editSet
@@ -75,15 +77,32 @@ class EditSetControllerSpec extends BaseControllerSpec {
         )
 
       status(editSet) mustBe OK
-      contentType(editSet) mustBe Some("text/html")
+    }
+
+    "redirect to the login page from the application when requested with invalid session token" in {
+      val mockEditSet = mock[editSet]
+      val mockEditSetService = mock[EditSetService]
+      val controller = new EditSetController(
+        Helpers.stubMessagesControllerComponents(),
+        mockEditSetService,
+        mockEditSet
+      )
+      val editSet = controller
+        .view("1")
+        .apply(
+          CSRFTokenHelper.addCSRFToken(
+            FakeRequest(GET, "/edit-set/1").withSession(SessionKeys.token -> invalidSessionToken)
+          )
+        )
+
+      status(editSet) mustBe SEE_OTHER
     }
 
     "order records" when {
 
-      def orderingRequest(field: String, direction: String, offset: Int = 1) = {
+      def orderingRequest(mockEditSet: editSet, field: String, direction: String, offset: Int = 1) = {
         val request = FakeRequest(GET, s"/edit-set/1?field=$field&direction=$direction&offset=$offset")
           .withSession(SessionKeys.token -> validSessionToken)
-        val mockEditSet = mock[editSet]
         val mockEditSetService = mock[EditSetService]
         val controller = new EditSetController(
           Helpers.stubMessagesControllerComponents(),
@@ -92,7 +111,7 @@ class EditSetControllerSpec extends BaseControllerSpec {
         )
         when(mockEditSetService.get("1")).thenReturn(IO.pure(getEditSet()))
         when(
-          mockEditSet(any[User], anyString(), anyString(), any[EditSetPage], any[EditSetEntryRowOrder])(
+          mockEditSet(any[User], anyString, anyString, any[EditSetPage], any[EditSetEntryRowOrder])(
             any[Messages],
             any[Request[AnyContent]]
           )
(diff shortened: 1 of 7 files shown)


test/support/BasePlaySpec.scala
```diff
@@ -0,0 +1,73 @@
...
+// * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+// * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+// * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+// * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+// * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// */
+
+package support
+
+import org.jsoup.Jsoup
```
tna-erasmos, line 45: Imports should be optimised.

test/support/BasePlaySpec.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.scalatest.{ BeforeAndAfterAll, BeforeAndAfterEach }
+import org.scalatest.matchers.must.Matchers
+import org.scalatestplus.play.PlaySpec
+import play.api.mvc.Results
+import play.api.test.Helpers.{ contentAsString, defaultAwaitTimeout }
+import play.twirl.api.Content
+import uk.gov.nationalarchives.omega.editorial.models.{ EditSet, User }
+import uk.gov.nationalarchives.omega.editorial.models.session.Session
+
+class BasePlaySpec
```
tna-erasmos, line 57: The `BaseISpec` is also a type of Play Spec; how about renaming this to `BaseControllerSpec`?

test/support/BasePlaySpec.scala
```diff
@@ -0,0 +1,73 @@
...
+  val messages: Map[String, Map[String, String]] =
+    Map(
+      "en" -> Map(
+        "edit-set.heading"       -> "Edit set: COAL 80 Sample",
+        "edit-set.table-caption" -> "Showing {0} - {1} of {2} records"
+      )
+    )
+
+  val validSessionToken: String = Session.generateToken("1234")
+  val mockEditSetView: EditSet = mock[EditSet]
```
tna-erasmos, line 69: But this isn't a view; perhaps you meant `mockEditSet`?

Then again - why mock the `EditSet`?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -69,846 +66,89 @@ class EditSetControllerSpec extends BaseSpec {
...
           CSRFTokenHelper.addCSRFToken(
-            FakeRequest(GET, "/edit-set/1")
-              .withSession(SessionKeys.token -> validSessionToken)
+            FakeRequest(GET, "/edit-set/1").withSession(SessionKeys.token -> validSessionToken)
           )
         )
 
       status(editSet) mustBe OK
       contentType(editSet) mustBe Some("text/html")
-      val document = asDocument(editSet)
```
tna-erasmos, line 103: Why have the content assertions been removed?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -69,846 +66,89 @@ class EditSetControllerSpec extends BaseSpec {
...
-                scopeAndContents =
-                  "Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths. (I)",
-                coveringDates = "1971"
-              )
-            ),
-            numberOfPages = 2
-          )
-        )
-
+        contentType(page) mustBe Some("text/html")
```
tna-erasmos, line 136: Why have the content assertions been removed?

tna-erasmos (commented):
This doesn't seem to follow the intent of the ticket, in that we should stick close to [this guidance](https://www.playframework.com/documentation/2.8.x/ScalaTestingWithScalaTest#Unit-Testing-Controllers).

I would have interpreted this task as making a lighterweight version of the corresponding integration test (`EditSetISpec`). However, as it is, it seems to have even less value than before.

I won't comment as to the value of having this test as well as the integration test. But I do think that we should make it as valuable as we can.

You might even want to base it on `EditSetISpec`.

test/support/BasePlaySpec.scala
userjd: Done

test/support/BasePlaySpec.scala
userjd: Removed

test/support/BasePlaySpec.scala
userjd: Done

test/controllers/EditSetControllerSpec.scala
userjd: Since the unit test returns some mock `HTML` data, the assertion is limited to the content type and the status of the request

rwalpole:
@tna-erasmos I probably should have read that [Unit Testing Controllers](https://www.playframework.com/documentation/2.8.x/ScalaTestingWithScalaTest#Unit-Testing-Controllers) page which I linked in the ticket a bit more carefully. In fact I _do not_ want us to follow their suggestion of using dependency injection for mocks for the simple reason that once you do this it ceases to be a unit test. By using dependency injection you are inevitably starting up the entire application in the background, which I know from painful experience can lead to unexpected and difficult-to-debug test failures. The intention of this ticket is to create simple unit tests that allow us to test individual components in isolation and obtain the assurance that they are working as expected.

tna-erasmos:
@rwalpole , if we're not following the approach described in [Unit Testing Controllers](https://www.playframework.com/documentation/2.8.x/ScalaTestingWithScalaTest#Unit-Testing-Controllers), then do you have an example of a controller-level test which is a unit test?

test/support/BasePlaySpec.scala
tna-erasmos: Confirmed.

test/support/BasePlaySpec.scala
tna-erasmos: Confirmed.

test/support/BasePlaySpec.scala
tna-erasmos: Confirmed that this instance was removed.

However, we're mocking the `EditSet` elsewhere (in `EditSetControllerSpec`) and I'm not sure that's what we want

But resolving this comment.

test/controllers/EditSetControllerSpec.scala
tna-erasmos: Which begs the question (as we discussed): don't we want to allow Play to generate the actual view?

Otherwise, what are we really testing here?

test/controllers/EditSetControllerSpec.scala
tna-erasmos: As above.

rwalpole:
Sure @tna-erasmos here's one: https://github.com/hmrc/secure-message/blob/main/test/uk/gov/hmrc/securemessage/controllers/SecureMessageControllerSpec.scala

tna-erasmos:
Thanks for the discussion, @userjd and @rwalpole; I believe I now understand what has been requested:

1. We can include mocks, but _without relying on dependency injection_, so constructing with `mock()`.
2. We will keep the current scenarios, _without adding new ones._
3. We will only assert _the status code_, **not** the content, which will remain the responsibility of the related view spec (in this case, `EditSetViewSpec`).

test/controllers/EditSetControllerSpec.scala
```diff
@@ -70,846 +69,89 @@ class EditSetControllerSpec extends BaseSpec {
...
-                scopeAndContents =
-                  "Bedlington Colliery, Newcastle Upon Tyne. Photograph depicting: view of pithead baths. (E)",
-                coveringDates = "1963"
-              )
+        val mockMessagesApi = stubMessagesApi(messages)
+        val mockEditSet = mock[editSet]
+        val mockEditSetService = mock[EditSetService]
+        val stub = stubControllerComponents()
+        val controller = new EditSetController(
+          DefaultMessagesControllerComponents(
```
felixgb, line 104: can we use `Helpers.stubControllerComponents()` ? I think `Helpers` comes from `play.api.test`

test/controllers/EditSetControllerSpec.scala
felixgb: I agree - I'm a bit wary here of mocking out the views here, especially as we've not added anything to the `EditSetViewSpec` - I think it would be useful to test some aspects of the views in isolation without running the integration test suite.

In a way I see the HTML as the output of this controller and I would expect assertions to be made on the filled out template - so we can stub the data that the controller uses to fill out the HTML then make assertions on that.

test/controllers/EditSetControllerSpec.scala
userjd: Done

test/controllers/EditSetControllerSpec.scala
tna-erasmos: `EditSetViewSpec` has some assertions on the content (more than the `EditSetControllerSpec` but much less than in `EditSetISpec`); perhaps this can be addressed in [the next ticket](https://national-archives.atlassian.net/browse/PACT-793).

tna-erasmos: I understand now that @rwalpole wishes for all content assertion to be one in `EditSetViewSpec`; only the status should be asserted here.

However, I see that we're still asserting the content type. Should we?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -48,868 +52,81 @@ class EditSetControllerSpec extends BaseSpec {
...
-        Map(
-          "en" -> Map(
-            "edit-set.heading"       -> "Edit set: COAL 80 Sample",
-            "edit-set.table-caption" -> "Showing {0} - {1} of {2} records"
-          )
-        )
-      val mockMessagesApi = stubMessagesApi(messages)
-      val editSetInstance = inject[editSet]
-      val stub = stubControllerComponents()
+      val mockEditSet = mock[editSet]
```
tna-erasmos, line 55: I would have expected constructing a real instance of the view; like this:

```
        new editSet(any[User], anyString(), anyString(), any[EditSetPage], any[EditSetEntryRowOrder])(
          any[Messages],
          any[Request[AnyContent]]
        )
```

@rwalpole, I believe you mentioned this, this morning?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -48,868 +52,81 @@ class EditSetControllerSpec extends BaseSpec {
...
         .apply(
           CSRFTokenHelper.addCSRFToken(
-            FakeRequest(GET, "/edit-set/1")
-              .withSession(SessionKeys.token -> validSessionToken)
+            FakeRequest(GET, "/edit-set/1").withSession(SessionKeys.token -> validSessionToken)
           )
         )
 
       status(editSet) mustBe OK
       contentType(editSet) mustBe Some("text/html")
```
tna-erasmos, line 78: Do we want to assert the content type?

test/controllers/EditSetControllerSpec.scala
```diff
@@ -48,868 +52,81 @@ class EditSetControllerSpec extends BaseSpec {
...
 
     "order records" when {
 
       def orderingRequest(field: String, direction: String, offset: Int = 1) = {
         val request = FakeRequest(GET, s"/edit-set/1?field=$field&direction=$direction&offset=$offset")
           .withSession(SessionKeys.token -> validSessionToken)
-        route(app, request).get
-      }
-
-      "CCR, ascending" in {
```
tna-erasmos, line 165: I thought that @rwalpole asked for all the scenarios to be maintained.

But many have been deleted.

test/support/BaseControllerSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.scalatest.matchers.must.Matchers
+import org.scalatestplus.play.PlaySpec
+import play.api.mvc.Results
+import uk.gov.nationalarchives.omega.editorial.models.User
+import uk.gov.nationalarchives.omega.editorial.models.session.Session
+
+class BaseControllerSpec
+    extends PlaySpec with Results with Matchers with BeforeAndAfterAll with BeforeAndAfterEach with MockitoSugar {
+
+  val messages: Map[String, Map[String, String]] =
```
tna-erasmos, line 35: This is not used anywhere, now.

test/support/BaseControllerSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    Map(
+      "en" -> Map(
+        "edit-set.heading"       -> "Edit set: COAL 80 Sample",
+        "edit-set.table-caption" -> "Showing {0} - {1} of {2} records"
+      )
+    )
+  val validSessionToken: String = Session.generateToken("1234")
+  val user: User = User("dummy user")
```
tna-erasmos, line 43: Unused.

test/support/BaseControllerSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+  val invalidSessionToken: String = Session.generateToken("invalid-user")
```
tna-erasmos, line 44: Unused.

tna-erasmos (commented):
Still a few issues - at least according to my latest understanding of the intent.

test/controllers/EditSetControllerSpec.scala
rwalpole: @tna-erasmos @userjd I don't see any benefit in asserting the content type here as we are not setting it explicity in the controller, but rather it is being inferred by the framework.

test/controllers/EditSetControllerSpec.scala
rwalpole: @tna-erasmos No need. As mentioned above.

test/controllers/EditSetControllerSpec.scala
rwalpole: These sorting tests would still seem to have value as the sorting occurs within the controller.

test/controllers/EditSetControllerSpec.scala
rwalpole: @tna-erasmos I don't know if this would work exactly as the editSet view constructor expects GovukPagination and GovukTable components as well. A mock is fine here and we will worry about constructing a view in the EditSetViewSpec.

rwalpole (changes requested):

test/controllers/EditSetControllerSpec.scala
tna-erasmos: As @rwalpole wrote above, I agree that the content type assertions should be removed.

test/controllers/EditSetControllerSpec.scala
tna-erasmos: OK.

test/controllers/EditSetControllerSpec.scala
userjd: Removed

test/controllers/EditSetControllerSpec.scala
userjd: Removed

test/controllers/EditSetControllerSpec.scala
userjd: Removed

test/controllers/EditSetControllerSpec.scala
userjd: Added all the tests for sorting and the invalid session token but not w3c reccomendations due to mockeditSet

test/support/BaseControllerSpec.scala
userjd: Removed

test/support/BaseControllerSpec.scala
userjd: Removed

test/support/BaseControllerSpec.scala
userjd: Used for test

test/controllers/EditSetControllerSpec.scala
```diff
@@ -48,868 +52,81 @@ class EditSetControllerSpec extends BaseSpec {
...
-        )
+        verify(mockEditSetService).get(anyString())
 
+        editSet
       }
 
-      "Page 2 sorted by CCR, ascending" in {
-        val page = orderingRequest("ccr", "ascending", offset = 2)
+      "with request parameters" in {
+        val page = orderingRequest("ccr", "ascending")
```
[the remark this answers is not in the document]
userjd: Thanks @rwalpole all the tests for sorting now added

rwalpole (approved):

test/controllers/EditSetControllerSpec.scala

test/controllers/EditSetControllerSpec.scala

test/controllers/EditSetControllerSpec.scala
tna-erasmos: Confirmed.

test/support/BaseControllerSpec.scala
tna-erasmos: Confirmed.

test/support/BaseControllerSpec.scala
tna-erasmos: Confirmed.

test/support/BaseControllerSpec.scala
tna-erasmos: Confirmed.

test/controllers/EditSetControllerSpec.scala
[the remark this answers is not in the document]
tna-erasmos: Confirmed.

PR #96, opened 2023-03-08
PACT-793 Replace integration tests in EditSetViewSpec with unit tests

userjd:


test/views/EditSetViewSpec.scala
```diff
@@ -22,22 +22,26 @@
...
 import uk.gov.nationalarchives.omega.editorial.views.html.editSet
 
-class EditSetViewSpec extends BaseSpec {
+class EditSetViewSpec extends BaseViewSpec {
 
   "Edit set Html" should {
     "render the given title and heading" in {
-
-      val editSetInstance = inject[editSet]
+      val mockGovukPagination = mock[GovukPagination]
```
tna-erasmos, line 42: Does these have to be mocked? 

If it can be easily constructed, I'd say that's always preferable to a mock.

test/views/EditSetViewSpec.scala
```diff
@@ -49,12 +53,17 @@ class EditSetViewSpec extends BaseSpec {
         editSet.entries.size,
         1
       )
+
+      when(mockGovukTable.apply(Table(anyList))).thenReturn(getEditSetTable())
```
tna-erasmos, line 57: I had expected that we'd render the whole view, with data, and preferable using the actual messages.

This is an odd hybrid of real and fake.

I wonder if we could be more realistic, and yet still have a quick test?

Perhaps I'm unclear of the general intent here.

test/views/EditSetViewSpec.scala
```diff
@@ -49,12 +53,17 @@ class EditSetViewSpec extends BaseSpec {
...
+      when(mockGovukPagination.apply(Pagination())).thenReturn(any)
       val editSetHtml: Html = editSetInstance(user, title, heading, editSetPage, EditSetEntryRowOrder.defaultOrder)(
         Helpers.stubMessages(),
         CSRFTokenHelper.addCSRFToken(FakeRequest())
       )
 
-      val document = asDocument(editSetHtml)
+      contentType(editSetHtml) mustBe "text/html"
+      println(editSetHtml)
```
tna-erasmos, line 65: Was this just temporary, the `println`?

test/views/EditSetViewSpec.scala
```diff
@@ -49,12 +53,17 @@ class EditSetViewSpec extends BaseSpec {
...
+      val document = asDocument(Helpers.contentAsString(editSetHtml))
       document must haveTitle(title)
```
tna-erasmos, line 67: We are asserting some of the aspects of the generated view.

But not nearly to the extent we are in the respective integration test (`EditSetISpec`).

Should we be?

If not, what subset should we be asserting?

test/views/EditSetViewSpec.scala
```diff
@@ -134,17 +145,93 @@ class EditSetViewSpec extends BaseSpec {
       editSet.entries.size,
       1
     )
+    when(mockGovukTable.apply(Table(anyList))).thenReturn(any)
```
tna-erasmos, line 148: It's a bit odd to return `any`; isn't there something more realistic to return. Why doesn't it matter?

test/support/BaseViewSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package support
+
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import uk.gov.nationalarchives.omega.editorial.models.User
+
+class BaseViewSpec extends UnitTest {
```
tna-erasmos, line 28: I think this is a good addition.

tna-erasmos (commented):
I might be unclear on the general intent.

test/views/EditSetViewSpec.scala
userjd: Replaced mock objects with govuk component.

test/views/EditSetViewSpec.scala
userjd: Removed now as mock objects with govuk component.

test/views/EditSetViewSpec.scala
userjd: Removed

test/views/EditSetViewSpec.scala
userjd: The task in this PR is to replace the existing integration tests with unit tests. To add additional scenarios from `EditSetISpec` like ordering would mean testing the controller, which has already been done in the `EditSetControllerSpec`

rwalpole (approved):

test/views/EditSetViewSpec.scala
tna-erasmos: Confirmed.

tna-erasmos: However, in some case the val name still has a `mock` prefix.

test/views/EditSetViewSpec.scala
tna-erasmos: Confirmed.

test/views/EditSetViewSpec.scala
tna-erasmos: I hadn't meant that we are now missing scenarios - I understand that the current task is only to switch to a unit test - but rather possible assertions of the view's contents.

If you compare the assertions we do in `EditSetISpec`, such as [here](https://github.com/nationalarchives/ctd-omega-editorial-frontend/blob/c57298a4d0990e52c5b3daef0ef774f1353d9d6e/it/EditSetISpec.scala#L16-L16), you can see that there are fewer here.

But I'm not sure if this is outside of scope for this ticket.

@rwalpole  can decide that.

test/views/EditSetViewSpec.scala
tna-erasmos: Confirmed.

test/views/EditSetViewSpec.scala
tna-erasmos: Confirmed.

test/views/EditSetViewSpec.scala
tna-erasmos: Actually, now that I see that @rwalpole has approved this PR, he must be OK with it.

So resolving.

tna-erasmos (approved):
This seems to be in line with the ticket.

Only one small suggested renaming task.

test/views/EditSetViewSpec.scala
userjd: Well spotted, done.

PR #97, opened 2023-03-09
PACT-800 Replace integration tests in LogoutControllerSpec with unit tests

felixgb:


test/support/UnitTest.scala
```diff
@@ -22,12 +22,12 @@
...
 import org.mockito.scalatest.MockitoSugar
-import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach}
+import org.scalatest.{ BeforeAndAfterAll, BeforeAndAfterEach }
 import org.scalatest.matchers.must.Matchers
 import org.scalatestplus.play.PlaySpec
 import play.api.mvc.Results
 
-/**
-  * This trait should be the base class for all unit tests
+/** This trait should be the base class for all unit tests
```
tna-erasmos, line 30: Really? Even one like `DateRangeSpec`?

test/controllers/LogoutControllerSpec.scala
```diff
@@ -44,21 +44,25 @@ class LogoutControllerSpec extends BaseSpec {
...
         redirectLocation(result) mustBe Some(loginPagePath)
         session(result).get(SessionKeys.token) mustBe empty
-
       }
 
     }
 
   }
 
+  class LogoutTestCase() {
```
tna-erasmos, line 62: This might be of general value in the `BaseControllerSpec`.

test/controllers/LogoutControllerSpec.scala
```diff
@@ -44,21 +44,25 @@ class LogoutControllerSpec extends BaseSpec {
         session(result).get(SessionKeys.token) mustBe empty
 
       }
-      "when already logged out" in {
 
-        val controller: LogoutController = inject[LogoutController]
+      "when already logged out" in new LogoutTestCase {
```
tna-erasmos, line 48: Why the change in formatting? 

It seems inconsistent both with the style in this class, and elsewhere, where we don't have newlines between these scenario statements.

tna-erasmos (commented):
Minor points.

test/controllers/LogoutControllerSpec.scala
felixgb: Sorry, I don't understand the question - is it referring to new lines between the test scenarios?

test/support/UnitTest.scala
felixgb: I did not add this comment, however I do disagree with it as there are plenty of cases where we don't want all of those traits, like you say the date range spec.

tna-erasmos: I know; but this is the first time I'm seeing it.

Resolving, as it's out of scope.

test/controllers/LogoutControllerSpec.scala
felixgb: I view it as a context for this suite only. I am against adding things to the BaseControllerSpec because:
* I don't want to pollute the namespace for all tests
* I think that mocks will conflict if we set up mocks in the base controller class. Mockito is hard enough to reason about is. I think each test should be responsible for setting up the class that it's testing. In this case, no mocks are being set up, but I imagine they will for the other controllers.

tna-erasmos: OK.

tna-erasmos (approved):
Only minor formatting issues remain.

test/controllers/LogoutControllerSpec.scala
felixgb: I added the newline here because I think that test declarations as similar to function declarations - with a new line before and after the definition.

test/support/UnitTest.scala
felixgb: I think we should delete this @rwalpole

test/support/UnitTest.scala
```diff
@@ -22,12 +22,12 @@
...
   */
-trait UnitTest extends PlaySpec with Results with Matchers with BeforeAndAfterAll with BeforeAndAfterEach with MockitoSugar
+trait UnitTest
```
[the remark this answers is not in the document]
rwalpole: My bad. Thanks @felixgb

test/controllers/LogoutControllerSpec.scala
rwalpole: I leave it up to individuals how much white space they put in a file - within reason of course.

rwalpole (approved):

PR #101, opened 2023-03-14
PACT-802 EditSetRecordEditDiscardSpec unit tests

felixgb:


tna-erasmos:
There's been a merge today, so you may want to rebase, @felixgb. I tried it locally and no conflicts.

test/views/EditSetRecordEditDiscardSpec.scala
```diff
@@ -21,33 +21,31 @@
...
 import play.api.i18n.Messages
 import play.api.test.Helpers
-import play.twirl.api.Html
 import support.BaseSpec
 import support.CommonMatchers._
+import uk.gov.hmrc.govukfrontend.views.html.components.GovukNotificationBanner
 import uk.gov.nationalarchives.omega.editorial.models.PhysicalRecord
 import uk.gov.nationalarchives.omega.editorial.views.html.editSetRecordEditDiscard
 
 class EditSetRecordEditDiscardSpec extends BaseSpec {
```
tna-erasmos, line 32: Given all of the assertions on the contents, I'd suggest that this be renamed to `EditSetRecordEditDiscardViewSpec`.

test/views/EditSetRecordEditDiscardSpec.scala
```diff
@@ -21,33 +21,31 @@
...
+        title = "EditRecordTitleTest",
+        heading = "EditRecordHeadingTest",
+        oci = "EditRecordOciTest",
+        message = "Any changes have been discarded. Showing last saved version.",
+        recordType = Some(PhysicalRecord)
+      )
       val document = asDocument(confirmationEditSetRecordEditHtml)
-      document must haveTitle(title)
+
+      document must haveTitle("EditRecordTitleTest")
```
tna-erasmos, line 48: It seems that we're doing the exact same content assertions as we are in `EditSetRecordISpec`. 

However, we're doing it in a very different way there. Should we use the same approach?

test/views/EditSetRecordEditDiscardSpec.scala
```diff
@@ -57,35 +55,33 @@ class EditSetRecordEditDiscardSpec extends BaseSpec {
         )
       )
       document must haveBackLink("/edit-set/1/record/EditRecordOciTest/edit", "edit-set.record.save.back")
-
     }
 
-    "render the header" in {
-
-      val document = generateDocument()
+    "render the header" in new TestCase {
```
tna-erasmos, line 60: I believe that one of the implicit requirements of this ticket is to keep the exact scenarios, so this is likely out of scope, but: why do we have a separate scenario just to verify the header?

test/views/EditSetRecordEditDiscardSpec.scala
```diff
@@ -57,35 +55,33 @@ class EditSetRecordEditDiscardSpec extends BaseSpec {
...
       document must haveVisibleLogoutLink
       document must haveLogoutLinkLabel("header.logout")
       document must haveLogoutLink
 
     }
+
   }
 
-  private def generateDocument(): Document = {
+  class TestCase() {
```
tna-erasmos, line 81: I don't mind this pattern, but it isn't consistent with most of the other specs which use class level members. 

Should we adopt this approach, consistently? 

Let's remember that - even without this fixture - each scenario gets a new instance of the class, and thus members.

tna-erasmos (commented):
No real problems, just suggestions.

test/views/EditSetRecordEditDiscardSpec.scala
felixgb: I'm not sure, but I think it was added when the Logout link was added to sign in pages. Lots of the view specs have a "render the header" test.

test/views/EditSetRecordEditDiscardSpec.scala
felixgb: I think that's fine to have them top level, I'll make that change.

test/views/EditSetRecordEditDiscardSpec.scala
felixgb: Looking at `EditSetRecordISpec` I'm not sure I understand what the different approaches are. Do you mean get the response once, then make multiple assertions on them?

test/views/EditSetRecordEditDiscardSpec.scala
felixgb: Agreed, I've made that change.

tna-erasmos: Confirmed.

test/views/EditSetRecordEditDiscardSpec.scala
tna-erasmos: OK. Perhaps it will be reviewed later.

Irrelevant to this ticket.

test/views/EditSetRecordEditDiscardSpec.scala
tna-erasmos: Confirmed.

test/views/EditSetRecordEditDiscardSpec.scala
tna-erasmos: Mostly the usage of `ExpectedEditRecordDiscardPage`.

test/views/EditSetRecordEditDiscardViewSpec.scala
```diff
@@ -21,33 +21,35 @@
...
 import play.api.test.Helpers
-import play.twirl.api.Html
 import support.BaseSpec
 import support.CommonMatchers._
+import uk.gov.hmrc.govukfrontend.views.html.components.GovukNotificationBanner
 import uk.gov.nationalarchives.omega.editorial.models.PhysicalRecord
 import uk.gov.nationalarchives.omega.editorial.views.html.editSetRecordEditDiscard
 
-class EditSetRecordEditDiscardSpec extends BaseSpec {
+class EditSetRecordEditDiscardViewSpec extends BaseSpec {
```
userjd, line 33: Should this extend `BaseViewSpec` for consistency?

tna-erasmos: Good idea.

felixgb: I've made this change now.

userjd (approved):

PR #116, opened 2023-05-02
PACT-943: implementing service-based queue name mapping

rwalpole:


app/uk/gov/nationalarchives/omega/editorial/config/Config.scala
```diff
@@ -23,7 +23,9 @@ package uk.gov.nationalarchives.omega.editorial.config
 
 case class Config(
   broker: HostBrokerEndpoint,
-  credentials: UsernamePasswordCredentials
+  credentials: UsernamePasswordCredentials,
+  defaultRequestQueueName: String,
+  legalStatusRequestQueueName: Option[String] = None
```
userjd, line 28: I understand this allows to switch between the data sources to use either the frontend or the services layer. Does this mean there would be different queue name(s) for the request endpoints in the API services?

rwalpole: Yes, exactly. I created a diagram which I posted [here](https://national-archives.atlassian.net/wiki/spaces/PACT/pages/51773453/Service-based+Queue+Mapping). Hopefully that will make it a bit clearer.

conf/application.conf
```diff
@@ -19,3 +19,5 @@ credentials = {
   username = "?"
   password = "?"
 }
+
+default-request-queue-name = "PACS001_request"
```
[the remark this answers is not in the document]
rwalpole: I tried that but it doesn't work. I think because we are using [PureConfig which is based on Typesafe Config](https://pureconfig.github.io/) which uses HOCON then this applies: https://github.com/lightbend/config/blob/main/HOCON.md#hyphen-separated-vs-camelcase

userjd (approved):

PR #119, opened 2023-06-12
PACT-872 Modify request and response for places of deposit to match new API schema

userjd:


app/uk/gov/nationalarchives/omega/editorial/services/jms/StubData.scala
```diff
@@ -48,7 +48,7 @@ trait StubData {
       endDateMonth = "12",
       endDateYear = "1962",
       legalStatusID = "http://catalogue.nationalarchives.gov.uk/public-record",
-      placeOfDepositID = "1",
+      placeOfDepositID = "S2",
```
rwalpole, line 51: The identifier returned by the API services is actually a URI so it would be more realistic if these were URIs, e.g. `http://catalogue.nationalarchives.gov.uk/agent.S2`. You might want to put the URI base in a variable to save repeating it.

userjd: Done

PR #132, opened 2024-07-10
Correct title of README.md

rwalpole:
Title doesn't read right currently. It is an Editorial Frontend of the Pan-Archival Catalogue and not a catalogue of the Editorial Frontend.


adamretter (approved):
LGTM

